PR #23, opened 2017-08-29
added banner as an option now along with styleguide component

tomkrupka:
via - https://waynedotedu.tpondemand.com/entity/60662

image


resources/views/partials/menu-top.blade.php
```diff
@@ -5,13 +5,17 @@
...
+            <div class="small-12 columns position-relative">
                 <div class="vertical-centering title-area">
                     <h1>
-                        <a href="/">
-                            @if(config('app.sub_title') !== null)<span>{{ config('app.sub_title') }}</span>@endif
-                            {{ $site['title'] }}
-                        </a>
+                        <div>
+                            <a href="/">
+                                @if($site['id'] !== 1669)
```
nickdenardis: Was this area meant to be changed? 

The `@if(config('app.sub_title') !== null)` check should still work if there is going to be any change to this area.

resources/views/partials/menu-top.blade.php
```diff
@@ -30,6 +34,11 @@
                         </div>
                     </div>
                 </div>
+
+                @if(isset($banner) && $banner != false)
```
nickdenardis: If this is the trigger for the banner perhaps we add to the [`README.md`](https://github.com/waynestate/base-site/blob/master/README.md#developing-views) on how to trigger it from a controller?

robertvrabel: Its triggered via per page promotions in the `Banner` group defined in the `PromoRepository` -> https://github.com/waynestate/base-site/pull/23/files/bc7aacab6ce16520b316a1811729193728d11359#diff-6bd8c85afde6cd73327d6415a509969aR71

So we don't need to trigger it from a controller at all. I'm cool with expanding the documentation though and explaining how these global promotion groups work!

resources/views/partials/menu-top.blade.php
tomkrupka: No it wasn't I will fix thanks!!

resources/views/partials/menu-top.blade.php
nickdenardis: Sounds good. I think documenting all the inherent features of the overall template would be a good next step.

PR #29, opened 2017-09-14
Feature/hero text

robertvrabel:
Goals:

* Allow the excerpt/link fields from the hero promotional group display on top of a hero image.
* Only allow this functionality to a list of controllers (an array defined in `app/config`). Homepage is the only one allowed by default.
* Ability to change the default "View More" text through `app/config`.
* Ability for us to add excerpt and/or link to the promotional group for a site and have it work without adding any code to the website.
* Update the rotating arrow icon with the cleaner and larger version that is used on our newer websites

contained-text

contained-text-link


config/app.php
```diff
@@ -88,6 +88,21 @@
...
+    |
+    | This enables the hero image to have excerpt text and a view more link.
+    | You can also specify which controllers you want this to work on by
+    | adding to the array. By default only the homepage controller
+    | is allowed to have this functionality.
+    |
+    */
+    'hero_text_enabled' => true,
+    'hero_text_controllers' => ['HomepageController'],
+    'hero_text_more' => 'View More',
```
nickdenardis: Let's sentence case this to conform to university style. "View more"

robertvrabel: I'll get that change into the next feature branch I'm working on!

PR #38, opened 2017-10-17
redirect to / if they are trying to hit the URL of an old site that u…

chrispelzer:
Due to old sites using `index.php` as the homepage and switching to using the new base without the `.php` extension, going to www.example.wayne.edu/index.php would 404 instead of being the homepage.

This will check the `%{THE_REQUEST}` and determine if it has index.php to redirect it to / and not conflict with the core index.php

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
+sudo: false
+dist: trusty
 language: php
 cache:
+    yarn: true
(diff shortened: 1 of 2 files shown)


nickdenardis:
This will also help any Google indexed pages still with `index.php` in them.

PR #50, opened 2017-10-30
Feature/update nav color

breakdancingcat:
* Let's upgrade our current look to be more in line with Wayne.edu; I've changed the black menu top over to a dark green. Let me know what you think! 
base-site-menu-change1

### DIFF
--- a/resources/scss/_vars.scss
+++ b/resources/scss/_vars.scss
@@ -9,3 +9,7 @@ $menu-top-font-size: 16px;
 // Slick
 $slick-font-path: "../fonts/";
 $slick-loader-path: "/_resources/images/";
+
+// Colors
+$darker-green: #10312c;
+$lighter-yellow: #ffd969;
--- a/resources/scss/_vars.scss
+++ b/resources/scss/_vars.scss
@@ -2,9 +2,9 @@
 $menu-top-breakpoint: map-get($breakpoints, 'offcanvas');
 
 // Menu-top
-$menu-top-background: $black;
-$menu-height: 70px;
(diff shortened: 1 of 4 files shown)


nickdenardis:
I'm in favor of this change.

Before approving, can you post a pic of it on mobile?

tomkrupka:
I dig it...  Black is way to stark of a difference

breakdancingcat:
@nickdenardis  Here is the menu opened. I don't think we need the green background anymore, the underlines create great visibility to where you are in the menu. 

screencapture-base-site-wayne-dev-styleguide-menu-top-1509460029030
screencapture-base-site-wayne-dev-styleguide-menu-top-1509460806034

tomkrupka:
I like the green top-bar bg!! Looks good! :-)

I'm sorry but I don't agree with the underlines solution for the menu hierarchy. I think its confusing for the everyday user.  

I do agree the present base hierarchy colors/solution need to be revisited and made better.  I think there is a bug or some CSS that isn't working as intended.  

Just my opinion, great work!! :-)

nickdenardis:
Hmm.. Yea if we are going to change the off-canvas style (my bad for not comparing the live one) we may want to discuss internally first and look at a few options before deciding on one.

tomkrupka:
Just wanted to make sure we knew the latest accordion baked into foundation is pretty good now. https://foundation.zurb.com/sites/docs/accordion-menu.html

It also has the option of toggle arrows which keep the main nav item separate.  Would need styling for sure but just wanted to throw out this as an option and would be easier to maintain.

PR #54, opened 2017-11-08
Pull composer packages only with minimum compatibility for PHP 7.0.10

chrispelzer:
This will only pull packages that satisfy php 7.0.10


nickdenardis:
Shouldn't we be targetting `7.0.x`?

chrispelzer:
It's changed now to be minimum 7.0.10

PR #65, opened 2017-12-08
Hero image rotating

robertvrabel:
Allow the ability to configure which controllers can have the hero image rotate and what that limit should be.

### DIFF
--- a/app/Repositories/PromoRepository.php
+++ b/app/Repositories/PromoRepository.php
@@ -71,6 +71,10 @@ class PromoRepository implements DataRepositoryContract, PromoRepositoryContract
             'banner' => 'page_id:'.$data['page']['id'].'|first',
         ];
 
+        if (in_array($data['page']['controller'], config('app.hero_rotating_controllers'))) {
+            $group_config = str_replace('|limit:1', '|limit:'.config('app.hero_rotating_limit'), $group_config);
+        }
+
         // Return the parsed promotions
         return $this->parsePromos->parse($promos, $group_reference, $group_config);
     }
--- a/config/app.php
+++ b/config/app.php
@@ -116,6 +116,20 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Hero Rotating Enabled
+    |--------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


app/Repositories/PromoRepository.php
```diff
@@ -71,6 +71,10 @@ public function getRequestData(array $data)
             'banner' => 'page_id:'.$data['page']['id'].'|first',
         ];
 
+        if (in_array($data['page']['controller'], config('app.hero_rotating_controllers'))) {
+            $group_config = str_replace('|limit:1', '|limit:'.config('app.hero_rotating_limit'), $group_config);
```
nickdenardis: This feels a little assuming. 

Any way to do this without assuming the `$group_config` is `limit:1`?  Or force a replacement the whole string (if defined in the config)?

robertvrabel: Yeah I don't love repeating `limit:1` but below are the reasons I kept it this way. If you have a better suggestion over all lets iterate on it :)

1. Since you can now define in `config/app.php` the hero limit, there is not need to touch this code anyway.

2. If someone does change change the limit code on line 70 `'hero' => 'page_id:'.$data['page']['id'].'|randomize|limit:1',` then they are opting into having hero's rotate across the entire site, which makes the code below completely irrelevant anyway.

3. I had considered leaving the `|limit:1` off line 70 and doing an else like:

```
if (in_array($data['page']['controller'], config('app.hero_rotating_controllers'))) {
    $group_config .= '|limit:'.config('app.hero_rotating_limit');
} else {
    $group_config .= '|limit:1';
}
```

The reason i didn't like this is because if someone puts a limit on line 70, there are now two limits being set if you aren't paying attention to the code below. We could do extra checks to see if a limit is already set, but since that isn't the desired functionality I didn't want to purposely support doing that.

nickdenardis: Ok cool. Yea I can't think of a better way at this point either. 

I haven't encountered this situation yet, but when I do and a different way comes to mind, we'll address then 😄

PR #76, opened 2018-01-18
Add example figure/figcaption with alignments

robertvrabel:
screen shot 2018-01-18 at 8 48 20 am


nickdenardis:
Should there be an example of this element without the `class` on the `<figure>` container? Or would that never happen in the CMS?

PR #82, opened 2018-02-05
Feature/fake images

robertvrabel:
Replacing fake images http://placehold.it with a local way to serve up fake images. This way if you don't have an internet connection, you can still develop features that require images.

Features
* Blocks images from being created that are too large of a resolution
* Blocks images that are not on the same host (prevents hotlinking)
* Allows custom text on the image. By default it will show the image dimensions.
* Fallback to `imagestring` if True Type Fonts aren't enabled for GD Library

Example output:
screen shot 2018-02-07 at 2 07 35 pm


robertvrabel:
This is now updated to use the image-faker package which drastically reduces the code for this PR :)

PR #107, opened 2018-04-04
Feature/html5 validation

robertvrabel:
This should fix all validation issues on https://validator.w3.org/nu/?doc=https%3A%2F%2Fbase.wayne.edu%2F

The only one it doesn't is the newrelic injection of the script tag on our servers. Not sure we can change that though.

@breakdancingcat @tomkrupka 

According to this validator you cannot have a div tag inside an h1 element. So I split this out and put an h2 for the surtitle. Can you review that changes to verify its good?

screen shot 2018-04-04 at 10 04 44 am


resources/views/partials/menu-top.blade.php
```diff
@@ -6,17 +6,17 @@
...
-                <div class="vertical-centering title-area">
-                    <h1>
-                        <div>
-                            @if(config('app.surtitle') !== null)
-                                @if($site['parent']['id'] === null && config('app.surtitle_main_site_enabled') === true || $site['parent']['id'] !== null)
-                                    <a href="{{ config('app.surtitle_url') }}"><span class="surtitle">{{ config('app.surtitle') }}</span></a>
-                                @endif
+                <div class="title-area{{ config('app.surtitle') === null ? ' vertical-centering' : '' }}">
+                    @if(config('app.surtitle') !== null)
+                        <h2 class="surtitle">
```
nickdenardis: This feels weird to have an `h2` before the `h1`. Should be flipped so the "site title" is a child of the "surtitle" in the architecture of the page.

breakdancingcat:
I was thinking that we could go back to how it was but instead of having a DIV change it to a SPAN, span is allowed inside H1

resources/views/partials/menu-top.blade.php
robertvrabel: It did for me too, but all my research indicates the ordering of headings don't matter. Its the significance of the content inside which dictates what number the heading should be. 

In this case the surtitle is a "secondary heading" to compliment the main site title which is the H1. We could very well make it just a div with a class, but i was trying to be semantically correct since it is a heading.

Can you provide an example of what you mean by a child of?

robertvrabel:
@breakdancingcat Yea I actually tried that but we would have to change the span in CSS to be block level so the vertical alignment worked. It just felt odd to me to do that which lead me down the path of splitting them up independently.

resources/views/partials/menu-top.blade.php
nickdenardis: Because the 'surtitle' is the higher context (assumption that it is in most cases, details below), it should be higher in the heading order and the title of the site should be one lower in the heading order because it is described by the higher sibling header. 

Reference: https://webaim.org/techniques/semanticstructure/

In this case, we can style the headings how ever we want. 

But there are three scenarios that our site headings can fall into. 

The first is what is proposed here and not ideal (or perhaps in some situations it may be):

screen shot 2018-04-04 at 10 49 32 am

"Surtitle" is a child of "Wayne State University".

The second is possible when the relationship between the "Wayne State University", "surtitle" and "site title" are independent:

screen shot 2018-04-04 at 10 49 42 am

The third is what suspect most of our sites will fall into. Where "Wayne State University" is (or can be) independent and the "site title" is describing "surtitle".

screen shot 2018-04-04 at 10 49 52 am

Example of the third: https://bodybequest.med.wayne.edu/

breakdancingcat: @nickdenardis What do you think about reverting it back to how it was in the previous tag, and making the div into a span tag but making it block level? Rob pointed out that goes against the normal implementation of a span, which is why he wanted to split them up independently.

robertvrabel: @nickdenardis @breakdancingcat  Ok i'm cool with flipping them. The only question I have is if there is no surtitle should we dynamically swap the site title to `<h1>` or leave it as `<h2>` always?

This would be the code leaving it as a `<h2>` always.

```
@if(config('app.surtitle') !== null)
    <h1 class="surtitle">
        @if($site['parent']['id'] === null && config('app.surtitle_main_site_enabled') === true || $site['parent']['id'] !== null)
            <a href="{{ config('app.surtitle_url') }}">{{ config('app.surtitle') }}</a>
        @endif
    </h1>
@endif

<h2>
    <a href="/{{ $site['subsite-folder'] !== null ? rtrim($site['subsite-folder'], '/') : '' }}">{{ $site['title'] }}</a>
</h2>
```

nickdenardis: I'm thinking since the page title is an `h1` that we should dynamically flip the site title to an `h1` if there is no `surtitle`.

I believe that would also match the previous setup `h1` with `span`.

But I also can see how that area could be filled with too many conditionals then. Let's see how the code turns out.

nickdenardis: Also, looks like in Laravel 5.5 they introduced the `str_finish` helper.

https://stillat.com/blog/2017/11/30/laravel-55-string-helper-method-finish

I am thinking this line:
`/{{ $site['subsite-folder'] !== null ? rtrim($site['subsite-folder'], '/') : '' }}`
can be replaced with this:
`{{ str_finish($site['subsite-folder'], '/') }}`

Which may help the clarify of this area.

PR #109, opened 2018-04-05
HTML5 Validation (headings update)

robertvrabel:
Based on this conversation: https://github.com/waynestate/base-site/pull/107

* Reverse the ordering of headings and style the h1 and h2 the same. 
* Changing the display to table so it vertically center properly based on the vertical-centering utility class


robertvrabel:
@nickdenardis Let me know your thoughts, i'll leave this open until then. I think the code came out pretty clean and it was easy to share the styling between the two instances.

PR #111, opened 2018-04-05
Allow image-list component to have images or buttons displayed depend…

robertvrabel:
I need a vote on if we should rename the existing partial or make a copy and call it `image-button-list.blade.php` with this behavior. I'm ok with leaving it as is.

screen shot 2018-04-05 at 12 03 39 pm

### DIFF
--- a/factories/UnderMenu.php
+++ b/factories/UnderMenu.php
@@ -25,7 +25,7 @@ class UnderMenu implements FactoryContract
         for ($i = 1; $i <= $limit; $i++) {
             $data[$i] = [
                 'link' => $this->faker->url,
-                'relative_url' => '/styleguide/image/360x131',
+                'relative_url' => $this->faker->boolean === true ? '/styleguide/image/360x131' : '',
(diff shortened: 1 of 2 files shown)


chrispelzer:
Wont you lose the display order if its 2 different partials

robertvrabel:
@chrispelzer Yea i'm not suggesting to use both partials in that way for under-menu. The additional partial would be images + buttons, like you see in this PR. 

The question is should we preserve just the image-list and make a image-button-list or something

chrispelzer:
Since it's a component and not that necessary of a need for a specific name, I say we update the name to match what it does.

robertvrabel:
@chrispelzer I made some updates. I noticed the styleguide image-list-lazy example was broken since that always requires an image. So I have two factories now to handle that case.

nickdenardis:
Thanks for this. This is great.

Personally I would name something with `cta` but I can't think or anything that describes it better than what you have here. 😛

breakdancingcat:
I like CTA for the future

PR #140, opened 2018-05-25
Feature/remove youtube parse package

chrispelzer:
Looks like this package was needed at one point but since it's being installed with the waynestate/parse-promos package at https://github.com/waynestate/parse-promos/blob/2.0.0/composer.json#L26

It should be left up to the package to install it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,8 +25,7 @@
         "waynestate/parse-menu": "~1.2",
         "waynestate/parse-promos": "~2.0",
         "waynestate/waynestate-api": "~1.2",
-        "waynestate/image-faker": "~1.0",
-        "waynestate/parse-youtube-id": "~1.0"
+        "waynestate/image-faker": "~1.0"
     },
     "require-dev": {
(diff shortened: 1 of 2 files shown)


robertvrabel:
Yea I can go either way with this. I'm not sure there is a rule to follow for a situation like this. The reason I left this as a direct dependency was incase someone used it a `Repository` that had nothing to do with Promos, example being parsing a link from Formy.

The question comes down to do you hinge directly on another packages (ParsePromos) dependency (YoutubeParser) still at that point? Or do you move it as direct dependency incase that package's dependencies change. If its the later, my thought is no one is going to remember to add it so i figured i'd leave it.

nickdenardis:
I'm in the same boat, could go either way. 

The question in my head would be, if we eventually replace `ParsePromos` with another package, would things break?

At the moment no, but if we add additional functionality (using the YouTube parser directly) then it would.

chrispelzer:
Closing out since it's being included with #142

PR #147, opened 2018-06-01
hotfix/5.0.10

chrispelzer:
Remove max-width and set width to be auto on images within tables. Otherwise the table column with the images will collapse if there is no width set on a `<td>` element

Move img and figure above the table section due to the stylelint `no-descending-specifity` rule
```
128:5  ✖  Expected selector ".content img" to come before selector ".content table img"   no-descending-specificity
```

version bump to 5.0.10

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "base",
   "private": true,
-  "version": "5.0.9",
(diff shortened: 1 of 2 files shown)


robertvrabel:
👍

PR #156, opened 2018-06-27
Snapchat icon

breakdancingcat:
Adding the ability to link users to your Snapcode from the social footer.


breakdancingcat:
Hotfix 5.1.3 instead

PR #161, opened 2018-07-02
Feature/breakpoints

robertvrabel:
Going to leave this open for discussion for a bit. Since the last time we changed the breakpoint ranges, it seems we have left the small breakpoint too large (360 pixels in total from 420 to 780). This PR changes the range so each breakpoint has an even distribution of 156 pixels by introducing two new breakpoint sizes for finer control.

Likely a few of these breakpoints won't be used often by base itself, but the key is they are there for when you do need them.

I did a once over on every component / template and with minimal effort made a few of them look better with these new breakpoints. The rest naturally looked good which was great!

### DIFF
--- a/resources/scss/components/_row.scss
+++ b/resources/scss/components/_row.scss
@@ -1,3 +1,3 @@
 .row {
-    @apply .mx-auto .max-w-screen-xl;
+    @apply .mx-auto .max-w-screen-xxxl;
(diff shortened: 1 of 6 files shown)


nickdenardis:
I like this idea of them being even and available if needed. I yield to y'all for feedback though.

PR #166, opened 2018-07-17
If a flickr embed exists fetch the script to make it work

robertvrabel:


robertvrabel:
From a discussion we had, we're going to not do this and somehow allow the script code to go through with the regular content for the site. This way if the embed changes over time from someone copy/pasting it, we won't need to maintain this.

PR #175, opened 2018-08-03
Feature/hero contained

breakdancingcat:
Hero full width contained doesn't look right, I don't think it should be a default 
https://homecoming.wayne.edu

### DIFF
--- a/config/base.php
+++ b/config/base.php
@@ -73,7 +73,7 @@ return [
     | site if it is not found in the menu.
     |
     */
-    'hero_contained' => true,
+    'hero_contained' => false,
 
     /*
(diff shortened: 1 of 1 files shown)


robertvrabel:
@breakdancingcat I agree that doesn't look great, but the reason we have it contained by default is so when a homepage DOES have a left menu, the hero image is small and not 100% across the site. Since most quick sites need it that way, I think it should be contained by default.

For the case of homecoming, I would set it to not contained for only the homepage or something if that is desired and they don't have a menu for that page.

breakdancingcat:
Oh, I think I get it now.

PR #177, opened 2018-08-08
Feature/base feature command

robertvrabel:
This PR adds a command to create a new feature quickly. Instead of having to manually create all the files necessary to start a new template this automates it by calling this artisan command:

`php artisan base:feature Spotlight`

Where Spotlight is the feature name. Its suggested for it to be singular (not the plural Spotlights name) and to be CamelCased if more than one word. The result is the creation of these files automatically:

* app/Http/Controllers/SpotlightController.php
* app/Repositories/SpotlightRepository.php
* contracts/Repositories/SpotlightRepositoryContract.php
* factories/Spotlight.php
* resources/views/spotlights.blade.php
* styleguide/menu.json
* styleguide/Pages/Spotlights.php
* styleguide/Repositories/SpotlightRepository.php

You can then visit `/styleguide/spotlights` in the browser and see a page with an example bulled list of data coming from the factory.

**Note the new file** `styleguide/menu.json`. In order to automatically inject menu items I converted the PHP array to a JSON file instead.


chrispelzer:
This works for me. The only thing I can think of is that there are some situations where you don't need a Controller for the feature.

robertvrabel:
@chrispelzer Yeah, sometimes you don't need repositories or factories either. We could prompt and ask which files you want created. Lets see how it goes, it might be just as easy to delete the parts you don't want by looking at the diff after running it.

PR #184, opened 2018-08-21
Hotfix/5.5.0

chrispelzer:
Updates the Formy Parser package to 1.0.2 to remove the default PHPSESSID key and use ini_get to get the key that is set to `session.name` by the server.


nickdenardis:
@chrispelzer 
Can you make sure the right commits are included in this release?

PR #191, opened 2018-09-04
Short title in search box for context

nickdenardis:
The quickest way to get a 'short' title in the university header if both elements exist. 

There are a number of ways to get this data in the DOM to be used for the header, the `data` attribute was used here for simplicity.

Since the `placeholder` attribute isn't read by screen readers, we could also manipulate the hidden search label to provide that context to a non-sighted user.

Will create styleguide examples of the on and off state if this is the direction we will go in.

## Result

screen shot 2018-09-04 at 7 23 56 am

### DIFF
--- a/factories/Page.php
+++ b/factories/Page.php
@@ -26,7 +26,8 @@ class Page implements FactoryContract
             $data[$i] = [
                 'site' => [
                     'id' => 2,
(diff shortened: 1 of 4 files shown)


nickdenardis:
Closing. Moving forward with https://github.com/waynestate/base-site/pull/192

PR #192, opened 2018-09-04
Short title used for mobile header display and search context

nickdenardis:
Using the 'short-title' field for both the search context and to be displayed visually when the top bar shrinks.

This allows for "Housing & Residential Life" to be shortened to "Housing" when space is constrained.

Not a huge fan of the duplicate HTML in the header but it does work correctly on VoiceOver for Mac, haven't tested it on others. 

Since the `placeholder` attribute isn't read by screen readers, we could also manipulate the hidden search label to provide that context to a non-sighted user.

Will create styleguide examples of the on and off state if this is the direction we will go in.

## Result

search

### DIFF
--- a/factories/Page.php
+++ b/factories/Page.php
@@ -26,7 +26,8 @@ class Page implements FactoryContract
             $data[$i] = [
                 'site' => [
                     'id' => 2,
-                    'title' => 'Styleguide',
+                    'title' => 'Site style guide',
+                    'short-title' => 'Styleguide',
                     'keywords' => '',
                     'subsite-folder' => null,
                     'parent' => [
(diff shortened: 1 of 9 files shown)


robertvrabel:
This approach with having it replaced in the menu top title is good! Did you test having a surtitle along with it? I want to make sure having "School of Medicine" then a long/short name still works properly.

robertvrabel:
If you check out this page it shows the surtitle even on small screens:  https://base.wayne.edu/styleguide/header/title/double

Another example from a med school site is: https://cme.med.wayne.edu/

Are we wanting to change that functionality?

nickdenardis:
@robertvrabel 
You're right, just checked on `develop` the surtitle does show on mobile. That makes it easier.

No intent to change that functionality. I didn't check to verify before making changes to the code. 

Thanks for catching that 👍

nickdenardis:
This PR has been updated to include the surtitle on mobile and the search label now reflects the short name for screen reader user context.

## Result

search-3

PR #221, opened 2018-10-03
Allow upper and lowercase titles of svg names

robertvrabel:
strtolower the actual variable value, not the variable name that was passed in.

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -44,7 +44,7 @@ class AppServiceProvider extends ServiceProvider
                 }
             }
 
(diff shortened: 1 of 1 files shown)


nickdenardis:
I'm good with this but wondering if it is a spot we could utilize the `slug()` helper:
https://github.com/illuminate/support/blob/dea093618c000cb151d5d946c920c802b8d6e09b/helpers.php#L976
https://github.com/illuminate/support/blob/7c1bee0a38d1e27ecf4817d294061ff50c7be8e3/Str.php#L418

robertvrabel:
@nickdenardis Yea that would well here

PR #223, opened 2018-10-05
Skip to content fix

robertvrabel:
Per: https://github.com/waynestate/base-site/pull/222#discussion_r223024594

Add tabindex to the content landmark to make it focusable. Remove it from the previous landmark that we were trying to skip to.


robertvrabel:
🤦‍♂️  Committed the file

PR #240, opened 2018-10-17
Feature/table img

breakdancingcat:
1. Following up on a request to fix images in tables, I removed the offending code. 
    - After specifying our image code for small screen size in a previous PR, we no longer need to make up for the problems that code caused at other break points.
    - I added an image column in the styleguide. 
2. While I was working in our content css, I took this as an opportunity to convert our styles over to tailwind classes, which offers values that are close enough. 
3. I noticed we had a `.footer-content` class that I removed and instead scoped the link color change to `footer .content`. 
    - This is why `footer.blade` was edited.
    - I added a link to the FooterContact factory since we specifically account for this in .content. 
4. The update to `<h1>` through `<h6>` margins needed to be accounted for in the `menu-top.blade`. 
    - This works in our favor because the height is now consistent switching between `Header title single` and `Header title double`, there was a slight difference before. 
5. The 3 columns of list examples is now responsive on the styleguide page. 
6. The 'figure' element has been updated and moved to a component 
7. I removed excess right margin from `banner.blade`.
8. Two/three column code updated in the content area; columns weren't behaving correctly on smaller screens
9. Added a `Dark button` example.

content-compare

mobile-compare

figure

### DIFF
--- a/styleguide/Views/styleguide.blade.php
+++ b/styleguide/Views/styleguide.blade.php
@@ -184,6 +184,8 @@
         <h2>Buttons</h2>
 
         <a href="#standard-button" class="button">Standard button</a>
+        <br />
+        <a href="#standard-button" class="button bg-gradient-green text-white">Dark button</a>
         <a href="#expanded-button" class="button expanded">Expanded button</a>
         <br />
         <a href="#button-examples" class="button" onclick="document.querySelector('pre.button-examples').classList.toggle('hidden');">See button code</a>
@@ -191,6 +193,7 @@
         <pre class="button-examples hidden" style="background: #EAEAEA; margin-bottom: 10px; overflow: scroll;">
         {!! htmlspecialchars('
 <a href="#" class="button">Standard button</a>
+<a href="#" class="button bg-gradient-green text-white">Dark button</a>
 <a href="#" class="button expanded">Expanded button</a>
         ') !!}
         </pre>
--- a/styleguide/Views/styleguide.blade.php
+++ b/styleguide/Views/styleguide.blade.php
@@ -6,14 +6,14 @@
     <div class="content">
         <hr>
 
-        <h2>Two column example</h2>
+        <h2>Two column layout</h2>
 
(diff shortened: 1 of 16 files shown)


breakdancingcat:
I didn't have time to comb over the `figure` in `.content` but I'll do that soon/tomorrow.

nickdenardis:
@breakdancingcat 
Will the `figure` in `.content` be a separate PR or on this same one?

Wondering when is a good time to review this.

styleguide/Views/styleguide.blade.php
```diff
@@ -9,11 +9,11 @@
         <h2>Two column example</h2>
 
         <div class="row -mx-4 flex flex-wrap">
-            <div class="w-full sm:w-1/2 px-4">
+            <div class="w-full sm:w-1/2 px-4 content">
```
robertvrabel: Looks like this whole page is already wrapped in .content so we shouldnt need it here

```suggestion
            <div class="w-full sm:w-1/2 px-4">
```

styleguide/Views/styleguide.blade.php
```diff
@@ -9,11 +9,11 @@
...
                 <p>{{ $faker->paragraph }}</p>
             </div>
 
-            <div class="w-full sm:w-1/2 px-4">
+            <div class="w-full sm:w-1/2 px-4 content">
```
robertvrabel: Same as above I believe we can get rid of it

styleguide/Views/styleguide.blade.php
```diff
@@ -119,11 +121,13 @@
 
         <h2>Abbreviations</h2>
 
-        <p><abbr title="Wayne State University">WSU</abbr></p>
+        <div class="content">
```
robertvrabel: This one too might be unnecessary

robertvrabel:
Thanks for your work on this, looks great!

breakdancingcat:
@robertvrabel  @nickdenardis  Ready for re-review

PR #248, opened 2018-10-26
Release/5.11.1

Fw7424:
Version bump


nickdenardis:
@Fw7424 
Looks like when you pushed this up the 'tags' didn't go with it.

Can you push this up again but check the box to include tags?

If you're doing it with command line, it would be  
`git push origin --all && git push origin --tags`

PR #257, opened 2018-11-16
Feature/events icon

breakdancingcat:
## Events with icon
* Removed "mini events" language and replaced with "events listing"
* Grouping events by date
* Added events icon

screencapture-base-wayne-local-styleguide-homepage-2018-11-16-17_54_34


breakdancingcat:
@nickdenardis @robertvrabel @chrispelzer 
I need help:
1. In the factory, I had to comment out a line. My tests weren't passing but I'm not exactly sure what I need to do since I'm grouping by random date. 
2. In the events repository, is there a way where we don't have to write an if !empty?

factories/Event.php
```diff
@@ -23,15 +23,15 @@ public function __construct(Factory $faker)
...
+            $data[$this->faker->dateTimeThisMonth('now')->format('Y-m-d')][$i] = [
                 'url' => $this->faker->url,
                 'title' => $this->faker->sentence(rand(6, 10)),
                 'date' => $this->faker->dateTimeThisMonth('now')->format('Y-m-d'),
                 'start_time' => $this->faker->dateTimeThisMonth('now')->format('H:i:s'),
                 'is_all_day' => $this->faker->boolean,
             ];
 
-            $data[$i] = array_replace_recursive($data[$i], $options);
+            //$data[$i] = array_replace_recursive($data[$i], $options);
```
nickdenardis: If this is no longer needed, let's remove the line completely.

breakdancingcat: I wrote in my comment "In the factory, I had to comment out a line. My tests weren't passing but I'm not exactly sure what I need to do since I'm grouping by random date." 

It probably needs to be here, I'm just not sure how to handle it.

chrispelzer: I fixed it to allow options to be passed in and the tests pass with it uncommented.

PR #260, opened 2018-12-10
Validate HTML for all styleguide pages

nickdenardis:
## Reason for change

The first step to adding HTML and accessibility validation (with https://github.com/pa11y/pa11y) to the test suite. 

## Things to note

- Initial release, this could be refactored more before releasing into master
- There are errors in the HTML now, will fix those before it is an automated test
- [Table `summary` deprecated in HTML5](https://www.w3.org/WAI/tutorials/tables/caption-summary/), switched to `caption`. The CMS CkEditor support both, we should probably remove the Summary option from that dialog.
- Added missing `</div>` in the figure styleguide page
- Still working on getting the output to mirror phpunit which has the graph first then all the errors after. 

## Checklist

- [x] Read through the code diff to ensure accuracy
- [x] SiteImprove accessibility check
- [ ] ~~Verified each page looks good on each viewport size~~
- [ ] ~~Added an example to the styleguide if applicable~~
- [x] Added tests to prove my code works
- [ ] ~~Refactored foreach loops into more maintainable collection chaining~~
- [x] Screen shot or video

## Screenshot of test run

screen shot 2018-12-10 at 7 16 02 am


robertvrabel:
It looks like the package uses http://validator.w3.org/nu . Does this mean an active internet connection is required to run the tests? If so, I don't know if we should automate that test with the pre-commit hook since all the current tests are able to run without an internet connection.

I'm in favor of adding it though, but maybe we can look at running it on deploy instead for this case? Let me know what you think!

nickdenardis:
@robertvrabel It does rely on the third party service. Running on deploy would be a good idea. 

I wish we had a way to report to the developer the errors that are returned though, even if we don't have it fail to deploy.

We would for sure need to gt all the pages validated before integrating this into CI though.

nickdenardis:
Will re-open at another time with just the automated tests.

The HTML changes have been pushed in with #264

PR #282, opened 2019-02-19
Feature/grid template

breakdancingcat:
Proposing an idea to have a controller that displays a grid of promotion items.
The data utilizes a custom page field "grid_promo_group_id" with a limit of 75 of the most recent items.


app/Repositories/GridRepository.php
```diff
@@ -0,0 +1,64 @@
...
+     * {@inheritdoc}
+     */
+    public function getGridPromos()
+    {
+        // If there is an grid custom page field then inject it into the group reference
+        if (!empty($data['data']['grid_promo_group_id']) && ! array_key_exists($data['data']['grid_promo_group_id'], $group_reference)) {
+            $group_reference[$data['data']['grid_promo_group_id']] = 'grid_promos';
+        }
+
+        $group_reference = [];
```
nickdenardis: Is this overwriting the `$group_reference` assignment that is happening directly above?

app/Repositories/GridRepository.php
```diff
@@ -0,0 +1,64 @@
...
+
+        $group_config = [
+            'grid_promos' => 'limit:75',
```
nickdenardis: Thoughts on this being configurable in the function call? 

```
public function getGridPromos($limit = 75)
{
...

$group_config = [
  'grid_promos' => 'limit:' . $limit,

...
```

factories/Grid.php
```diff
@@ -0,0 +1,42 @@
...
+     * {@inheritdoc}
+     */
+    public function create($limit = 1, $flatten = false, $options = [])
+    {
+        for ($i = 1; $i <= $limit; $i++) {
+            $data[$i] = [
+                'relative_url' => '/styleguide/image/550x400?text=550x400%20',
+                'title' => $this->faker->text(20),
+                'link' => $this->faker->url,
+                'description' => '<p>' . $this->faker->text(5) . ' <a href="#">'. $this->faker->sentence(3)  .'</a>',
```
nickdenardis: Can we make this link a fake link instead of a hash? Looks like this also needs a closing paragraph tag.

resources/views/components/promo-grid.blade.php
```diff
@@ -0,0 +1,18 @@
...
+<div class="flex flex-wrap -mx-4">
+    @foreach($promo_grid as $item)
+        <div class="w-full md:w-1/2 xl:w-1/3 px-4 pb-6">
+            <a href="{{ $item['link'] }}" class="flex md:block">
+                <div class="w-1/3 md:w-full">
+                    <img src="{{ $item['relative_url'] }}" alt="" role="presentation" />
+                </div>
+                <div class="w-2/3 pl-4 md:pl-0">
+                    <div class="font-bold hover:underline">{{ $item['title'] }}</div>
+                    <p class="text-sm">{!! strip_tags($item['description']) !!}</p>
```
nickdenardis: Since this is stripping tags from the description is it even worth using description? Perhaps `excerpt` may be more appropriate.

resources/views/grid.blade.php
```diff
@@ -0,0 +1,28 @@
...
+
+@section('content')
+    @include('components.page-title', ['title' => $page['title']])
+
+    <div class="content">
+        {!! $page['content']['main'] !!}
+    </div>
+
+    @if(!empty($grid))
+        <div class="flex flex-wrap -mx-4">
```
nickdenardis: How is this block different than `components/promo-grid.blade.php`?

styleguide/Views/styleguide-grid.blade.php
```diff
@@ -0,0 +1,28 @@
+@extends('components.content-area')
```
nickdenardis: Thinking this should pick up the main `resources/views/grid.blade.php` instead of being a replica of the code.

resources/views/components/promo-grid.blade.php
```diff
@@ -0,0 +1,18 @@
+{{--
+    $items => array // [['title', 'description']]
+--}}
+<div class="flex flex-wrap -mx-4">
+    @foreach($promo_grid as $item)
+        <div class="w-full md:w-1/2 xl:w-1/3 px-4 pb-6">
```
nickdenardis: Thoughts on marking this up as a list?

app/Repositories/GridRepository.php
```diff
@@ -0,0 +1,64 @@
...
+        $this->cache = $cache;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getGridPromos()
+    {
+        // If there is an grid custom page field then inject it into the group reference
+        if (!empty($data['data']['grid_promo_group_id']) && ! array_key_exists($data['data']['grid_promo_group_id'], $group_reference)) {
```
robertvrabel: It doesn't look like `$data` is being passed as a parameter into this method so it will always result in this condition being false. Also looks like `group_reference` is being referenced but not defined above it.

app/Repositories/PromoRepository.php
```diff
@@ -96,6 +96,11 @@ public function getRequestData(array $data)
             $group_reference[$data['data']['accordion_promo_group_id']] = 'accordion_page';
         }
 
+        // If there is an grid custom page field then inject it into the group reference
```
robertvrabel: Since you have a `GridController` is this necessary to check on every request? The reason accordion is here is because it allows the `ChildPageController` to have an accordion at the bottom of every page. It seems like this feature you have to select the `GridController` anyway so it should be unnecessary the grid globally.

styleguide/Repositories/PromoRepository.php
```diff
@@ -53,6 +53,14 @@ public function getRequestData(array $data)
         // Only pull accordion for childpage template
         $accordion = !empty($accordion_page_ids[$data['page']['id']]) ? app('Factories\Accordion')->create($accordion_page_ids[$data['page']['id']]) : null;
 
+        $grid_page_ids = [
```
robertvrabel: In reference to the comment above about having to select the `GridController` anyway, all of the changes to this file shouldn't be necessary since they are handled by the `GridController` and `GridRepository`. The styleguide page you added `styleguide/Pages/Grid.php` already points to the `GridController` so that should be good to go!

resources/views/grid.blade.php
breakdancingcat: This was a duplicate, thanks for catching

robertvrabel:
To resolve the coverage/coveralls decrease we'll have to write some new tests. Once you have gone through all the comments and everything is good to go let's go over that together so I can show you how to run the coverage locally and add tests :)

app/Repositories/GridRepository.php
breakdancingcat: Ok, added this.

factories/Grid.php
breakdancingcat: Got it

resources/views/components/promo-grid.blade.php
breakdancingcat: Yep, I'll change this to excerpt. Thanks!

app/Repositories/PromoRepository.php
breakdancingcat: This was leftover from an idea I had to include it like accordion that I went back on. Removing!

app/Repositories/GridRepository.php
breakdancingcat: I moved the empty group reference above this and made sure data was being passed in

breakdancingcat:
@nickdenardis @robertvrabel  I went through all the comments and fixed everything! I'm good to sit down and write some tests.

breakdancingcat:
@nickdenardis 
@robertvrabel trained me on writing tests for my code and I think this template is ready to be merged if you agree!

breakdancingcat:
Screen shots! 
grid-desktop
grid-medium
grid-mobile

styleguide/Repositories/GridRepository.php
```diff
@@ -0,0 +1,19 @@
...
+
+class GridRepository extends Repository
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getGridPromos(array $data, $limit = 75)
+    {
+        return [
+            // Contact footer
```
robertvrabel: This comment can be removed

app/Http/Controllers/GridController.php
```diff
@@ -0,0 +1,38 @@
+<?php
+/*
+* Status: Public
+* Description: Grid Template
```
robertvrabel: Thoughts on a more descriptive description? Maybe `Grid Promotion Template` to say it's coming from promotions? Trying to think from the perspective of someone selecting it from the dropdown in the CMS.

app/Http/Controllers/GridController.php
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Display the grid view.
+     *
+     * @param Request $request
+     * @return \Illuminate\View\View
+     */
+    public function index(Request $request)
+    {
+        // Get grid promotions
```
robertvrabel: You can remove this comment `// Get grid promotions`. I'm trying to reduce comments that are clearly explained by well named code

robertvrabel:
@breakdancingcat Couple really small changes I commented on. Then I'm good for you to merge!

PR #298, opened 2019-03-27
Feature/global data

robertvrabel:
### Reason for change

This changes the structure of the global promos config. The way it is structured now is very limiting. An example is if you want a feature like "Announcements" from a promo group across the entire site + the subsites. Currently, when you goto the subsite it will pull in the config for that subsite if one is set but it won't pull additional promo data from the main site anymore. Currently the contact footer and social footer are the only ones that are always pulled.

This proposal changes the name of "Main" to "All" to imply that all this data will pulled across all sites. Then you can create an array under the "Sites" key to specify overrides. It even allows you to now specify the main site id under "Sites" to configure things specifically for the main site.

Additionaly, this opened up the opportunity to allow for more than configuring only promo data. A new key called "callbacks" exists for "All" and every subsite. This allows you to specify a string in the format of `Class@Method` and that will automatically be called for you and passed down to the view. This allows us to pull non promo driven data across the entire website easily.

Currently the two pieces of data that go across the whole website are:

```
            'callbacks' => [
                '\Repositories\MenuRepository@getRequestData',
                '\Repositories\PromoRepository@getRequestData',
            ],
```

I also added more test converage for specific cases.


breakdancingcat:
Awesome! @robertvrabel  What do we do if we want to change the config of a promo group in the parent site as it appears on the sub site? EG I want to use "spotlights" but I want my config to be different.

robertvrabel:
@breakdancingcat Under the "Sites" key you can specify the main site id as the key:

```
                'spotlights' => [
                    'id' => 4246,
                    'config' => 'page_id:{$page_id}|first', // update this to have a different config
                ],
```

This would override the what was set in "All". You would have to replicate the promo group id though. We could explore the ability to only specify the config and override it later though!

config/base.php
```diff
@@ -193,45 +193,67 @@
 
     /*
     |--------------------------------------------------------------------------
-    | Global Promotion Groups
+    | Global Data
```
chrispelzer: Wasn't this going to be under it's own globaldata config?

edit: nevermind, for some reason, the review area is showing only certain commits

PR #303, opened 2019-04-11
Accessibility - Tabbing left menu fix

robertvrabel:
Screen readers will read the `<hr>` tag at the end of the menu which isn't ideal. Since this is used for spacing only, I moved it to CSS instead which makes it look identical.

This PR introduced it: https://github.com/waynestate/base-site/commit/f49fd3af240ce8bc38dfc18573ba01f17ee0f5e0#diff-5af06945f27d37e67539117206b015a3R27

### DIFF
--- a/resources/scss/components/_menu-main.scss
+++ b/resources/scss/components/_menu-main.scss
@@ -1,6 +1,6 @@
 .main-menu {
     ul {
-        @apply .list-reset .m-0 .p-0;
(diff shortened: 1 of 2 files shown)


nickdenardis:
Does this visually look different? Or was the HR hidden to begin with?

robertvrabel:
@nickdenardis No visible difference. The HR wasn't due to a border 0 on it

PR #305, opened 2019-04-23
Add optional heading classes for components

robertvrabel:
This allows you to customize the component header without having to edit the file directly.

### DIFF
--- a/resources/views/components/article-listing.blade.php
+++ b/resources/views/components/article-listing.blade.php
@@ -4,7 +4,7 @@
     $url => string '/news/'
     $link_text => string // 'More news'
 --}}
-<h2>{{ $heading ?? 'News' }}</h2>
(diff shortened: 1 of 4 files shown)


resources/views/components/article-listing.blade.php
```diff
@@ -4,7 +4,7 @@
     $url => string '/news/'
     $link_text => string // 'More news'
 --}}
-<h2>{{ $heading ?? 'News' }}</h2>
+<h2{!! !empty($class) ? ' class="'.$class.'"' : '' !!}>{{ $heading ?? 'News' }}</h2>
```
nickdenardis: Can we add this to the documentation on the top of the file for each of these?

robertvrabel: yep! https://github.com/waynestate/base-site/pull/306

PR #309, opened 2019-04-24
Append PDF text to links for accessibility

robertvrabel:
Example "Download" becomes "Download (pdf)":

Screen Shot 2019-04-24 at 11 56 44 AM

### DIFF
--- a/resources/js/modules/pdf.js
+++ b/resources/js/modules/pdf.js
@@ -1,0 +1,9 @@
+import 'accordion/src/accordion.js';
+
+(function() {
+    "use strict";
+
+    document.querySelectorAll('a[href$=".pdf"]').forEach(function(item) {
+        item.innerHTML += ' (pdf)';
+    });
+})();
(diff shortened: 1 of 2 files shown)


resources/js/modules/pdf.js
```diff
@@ -0,0 +1,9 @@
+import 'accordion/src/accordion.js';
```
nickdenardis: Is this needed?

nickdenardis:
For reference this (pdf) does get read by NVDA:

targetprocess-screen-capture

resources/js/modules/pdf.js
robertvrabel: It is not, I'll remove!

robertvrabel:
@nickdenardis Yea we can add it to the styleguide, here is the PR. Let me know your thoughts on the wording. https://github.com/waynestate/base-site/pull/311

PR #317, opened 2019-05-02
Feature/slideout accessibility

robertvrabel:
*Trap the focus in the slideout when they open it. Once they tab through everything it should cycle back to the close (X) button again (not go into the content-area). The user needs to close the offcanvas themselve in order to get back to the content-area.
* Make sure it says "expanded" and "closed" when using the button AND when you toggle the "Main Menu"
* Make sure it does not read "Greater than" for the icon toggle on "Main Menu" for screen readers.
* Make sure you can tab onto and open "Main Menu" in the slideout

### DIFF
--- a/resources/views/components/content-area.blade.php
+++ b/resources/views/components/content-area.blade.php
@@ -16,7 +16,7 @@
                     <div class="offcanvas-main-menu mt:hidden">
                         <ul>
                             <li>
-                                <a class="main-menu-toggle">Main Menu</a>
+                                <a class="main-menu-toggle" tabindex="0">Main Menu</a>
 
                                 {!! $top_menu_output !!}
                             </li>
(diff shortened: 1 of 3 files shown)


resources/views/components/content-area.blade.php
```diff
@@ -16,7 +16,7 @@
                     <div class="offcanvas-main-menu mt:hidden">
                         <ul>
                             <li>
-                                <a class="main-menu-toggle">Main Menu</a>
+                                <a class="main-menu-toggle" tabindex="0" aria-expanded="false">Main Menu</a>
```
robertvrabel: I had to put tabindex="0" in order to be able to tab to this link. I am not sure why the DOM didn't naturally pick it up, but this fixes the current issue of not even being able to tab to this element.

PR #320, opened 2019-05-02
Firefox Focusring Accessibility

robertvrabel:
Firefox uses the text color as the focus ring. This causes issues of the text color is white and the overall container background is white. This forces the ring to be black in this case (to match the default of most cases of being the best contrast against white).


breakdancingcat:
Thanks!

PR #322, opened 2019-05-03
Feature/accordion accessibility updates

robertvrabel:
* Display none on actual content so you can’t tab into it if it is closed (if there is a link within that content).
* Fix for not reading the + and - icon for open and close with a screen reader
* Tabbing to the first tab in the accordion now says: "Collasped, tab list 5 items" instead of "Tab list 1 of 1"
* Updated the example accordion copy/paste code.

### DIFF
--- a/resources/views/components/accordion.blade.php
+++ b/resources/views/components/accordion.blade.php
@@ -4,7 +4,7 @@
 <ul class="accordion">
     @foreach($items as $key=>$item)
         <li>
-            <a href="#definition-{{ $key }}">{{ $item['title'] }}</a>
+            <a href="#definition-{{ $key }}"><span aria-hidden="true"></span>{{ $item['title'] }}</a>
             <div class="content" id="definition-{{ $key }}">{!! $item['description'] !!}</div>
         </li>
     @endforeach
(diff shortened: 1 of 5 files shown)


resources/views/components/accordion.blade.php
```diff
@@ -4,7 +4,7 @@
 <ul class="accordion">
     @foreach($items as $key=>$item)
         <li>
-            <a href="#definition-{{ $key }}">{{ $item['title'] }}</a>
+            <a href="#definition-{{ $key }}"><span aria-hidden="true"></span>{{ $item['title'] }}</a>
```
nickdenardis: Does this validate with this empty tag? Or are you filling it with something after load?

robertvrabel: @nickdenardis it is filled with a pseudo-element from CSS for ::content + or - so not sure if that counts? I could put something in there if we really need to, but I didn't want to add it since the content comes from CSS.

Are you talking about only HTML validation?

nickdenardis: Ok. We'll see how this all shakes out when running the tests on dev next week :)

PR #330, opened 2019-05-15
Feature/tailwind upgrade

robertvrabel:
This PR brings tailwind to the 1.0 release. There are no visual changes to the base website at all. Mainly the changes for us include:

* New class names for some of the utility classes
* Added/remove classes from html elements based on the new `@base` code (previously `@preflight`). The resets are slightly different.
* When using values from tailwind config inside our SCSS files the function is now `theme()` (previously `config()`
* The config file now includes only the changes we made to tailwind values


breakdancingcat:
Not too much of a difference, but their changes make sense. Thanks for making this PR so quickly!

PR #335, opened 2019-05-29
Feature/tailwind config spacing

breakdancingcat:
Tailwind config had 3/4 instead of 4/3 for spacing with a value of 75%. 3/4 would be 133.33% (portrait).

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -84,7 +84,7 @@ module.exports = {
                 'screen-xxxl': screens.xxxl,
             },
             padding: {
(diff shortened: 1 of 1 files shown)


nickdenardis:
Is there a before/after screenshot that shows what is wrong and now fixed?

nickdenardis:
Also, this feature branch PR should probably point to `develop`

breakdancingcat:
@nickdenardis I was mistaken, ratios are height:width, so 3/4 is correct.

PR #348, opened 2019-06-20
Hotfix/6.6.6

chrispelzer:
Update the permissions to be 02770 for the servers

Switch over the LOG_CHANNEL to be stack to now use both the daily log and passing the log onto the errorlog as well for use in AppOptics


breakdancingcat:
🤘

PR #358, opened 2019-07-02
Remove example buttons and expose the code so it is more accessible.

robertvrabel:
The buttons to toggle open/close the code are not accessible since they did not say "expanded/collapsed" with a screen reader. In order to make that work properly, more javascript code would need to be added to do the toggle and it doesn't make sense to do that all inline.

In an effort to not bloat the javascript file with styleguide specific code, I decided to remove the buttons entirely and expose the code all the time. It does make the page a lot longer, but I think it looks pretty good and I enjoy seeing the code exposed all the time for reference.

I'll leave this open for discussion since it is a larger change.

# Example:

Screen Shot 2019-07-02 at 11 10 41 AM

### DIFF
--- a/styleguide/Views/styleguide-accordion.blade.php
+++ b/styleguide/Views/styleguide-accordion.blade.php
@@ -11,9 +11,7 @@
         @include('components.accordion', ['items' => $accordion_page])
     @endif
 
-    <a href="#accordions" class="button" onclick="document.querySelector('pre.accordions').classList.toggle('hidden');">See accordion code</a>
-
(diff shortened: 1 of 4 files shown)


chrispelzer:
Works for me, though I've wanted to prettify that code chunk example for a while

robertvrabel:
@chrispelzer Would love to do that too, but again I don't want to bloat the site. We could have a separate CSS/JS file for only the styleguide but I am not sure I want to maintain that. Open to suggestions though!

PR #359, opened 2019-07-08
Add aria-labelledby for svgs that have links around them.

robertvrabel:
NVDA on windows does not read the labelledby on the svg, it requires it on the link element itself. Leaving it on both makes it work properly for NVDA and voice over on mac.

### DIFF
--- a/resources/views/components/breadcrumbs.blade.php
+++ b/resources/views/components/breadcrumbs.blade.php
@@ -6,7 +6,7 @@
         @foreach($breadcrumbs as $key=>$crumb)
             @if($key == 0)
                 <li class="inline">
-                    <a href="/"><span class="text-black align-middle">@svg('home', 'w-4 h-4 inline align-baseline')</span></a>
+                    <a href="/" aria-labelledby="Home"><span class="text-black align-middle">@svg('home', 'w-4 h-4 inline align-baseline')</span></a>
                     <span class="icon-right-open px-2"></span>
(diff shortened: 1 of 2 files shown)


resources/views/components/breadcrumbs.blade.php
```diff
@@ -6,7 +6,7 @@
         @foreach($breadcrumbs as $key=>$crumb)
             @if($key == 0)
                 <li class="inline">
-                    <a href="/"><span class="text-black align-middle">@svg('home', 'w-4 h-4 inline align-baseline')</span></a>
+                    <a href="/" aria-labelledby="Home"><span class="text-black align-middle">@svg('home', 'w-4 h-4 inline align-baseline')</span></a>
```
nickdenardis: I believe this should be `aria-label` per https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute unless you're planning to point to an `id` on the page.

nickdenardis: Ah yes, that is correct. I didn't look into the SVG to see the `id` there. Looking though it is, `home` but the capitalization of `Home` threw me off.

Can you change this to match the ID? `aria-labelledby="home"`

robertvrabel: Will do in the next PR!

PR #374, opened 2019-08-01
hotfix/6.8.4

robertvrabel:
* Set the subsite path if it exists to the topic url
* Remove the limit of topics since it actually isn't possible to do that yet


nickdenardis:
Is there a way to build this into the tests?

robertvrabel:
@nickdenardis Yea I'll make a note to do that

PR #375, opened 2019-08-08
feature/meta

robertvrabel:
* Allow custom page fields to override meta information
* Change article meta override to include alt text
* Make twitter handle configurable
* Add twitter card alt text


app/Http/Middleware/Data.php
```diff
@@ -48,6 +48,12 @@ public function handle($request, Closure $next)
             return $page;
         }
 
+        // Overrides for meta information
+        $data['meta']['image'] = !empty($page['data']['meta_image']) ? $page['data']['meta_image'] : '';
```
nickdenardis: Thoughts on changing some of these long conditionals to use `Array::get()`?
https://laravel.com/docs/5.8/helpers#method-array-get

Something like:
`Array::get($page, 'data.meta_image', '');`

PR #381, opened 2019-08-19
feature/wild-care-route

robertvrabel:
I noticed when running `php artisan route:cache` it would error since we are using a closure on the wild card route. This PR moves it to a proper Controller method so it is cacheable.

Error:
local.ERROR: Unable to prepare route [{path}] for serialization. Uses Closure. {"exception":"[object] (LogicException(code: 0): Unable to prepare route [{path}] for serialization. Uses Closure. at /vagrant/base/vendor/laravel/framework/src/Illuminate/Routing/Route.php:917)

### DIFF
--- a/Envoy.blade.php
+++ b/Envoy.blade.php
@@ -93,6 +93,7 @@ extractreleasepack_on_remote
 syncshareddata_remotesrc
 baseenvlink_remoterelease
 prepare_remoterelease
+cache_remote_release
 link_newrelease_on_remote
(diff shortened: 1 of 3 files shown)


robertvrabel:
I've updated this PR to include calling `config:cache` and `route:cache` on remote deployment since we weren't doing that in the past. This does require calling the alias on our servers `php73` so we will have to be mindful of updating that whenever we update the php version.

PR #382, opened 2019-08-21
Feature/accessibility audit

robertvrabel:
* Change all dynamic links to wayne.edu so we don't send users off to undesirable websites
* Add to the pre-commit hook an automatic run of the accessibility-checker: https://github.com/waynestate/accessibility-audit/pull/2/files . This will automatically scan every page found in the `styleguide/menu.json` file and report issues to you.
* Introduce `axe.config.json` which allows filter overrides for the accessibility-checker. This is mainly for color contrast issues regarding gradient backgrounds since they can't be automatically tested.


styleguide/Http/Controllers/UnderMenuController.php
```diff
@@ -68,14 +68,16 @@ public function index(Request $request)
             'secondary_alt_text' => 'Example dark icon',
         ]);
 
-        // Bg image light
+        // // Bg image light
```
nickdenardis: Small double comment here.

factories/HeroImage.php
```diff
@@ -26,7 +26,7 @@ public function create($limit = 1, $flatten = false, $options = [])
             $data[$i] = [
                 'relative_url' => '/styleguide/image/1600x580?text=1600x580%20('.$i.')',
                 'title' => $this->faker->sentence,
-                'description' => '<p>' . $this->faker->text(100) . ' <a href="#">'. $this->faker->sentence(3)  .'</a></p>',
+                'description' => '<p>' . $this->faker->text(100) . ' <a href="https://wayne.edu">'. $this->faker->sentence(3)  .'</a></p>',
```
nickdenardis: Perhaps we should have all wayne.edu URLs with or without the trailing slash for consistency?

PR #383, opened 2019-08-21
feature/deploy-cache

robertvrabel:
Properly call the alias and path. Tested this on Jenkins and verified they are called successfully now.

### DIFF
--- a/Envoy.blade.php
+++ b/Envoy.blade.php
@@ -264,11 +264,13 @@ echo "RemoteRelease Prepare Done.";
 @endtask
 
 @task('cache_remote_release', ['on' => $remote_server])
+shopt -s expand_aliases
+source ~/.bashrc
 echo "Caching configs...";
-php73 artisan config:cache
+php73 {{ $source_dir }}/ {{ $release_dir }}/{{ $release }}/artisan config:cache
 echo "Caching routes...";
-php73 artisan route:cache
-echo "Done caching";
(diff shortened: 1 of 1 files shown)


Envoy.blade.php
```diff
@@ -264,11 +264,13 @@
 @endtask
 
 @task('cache_remote_release', ['on' => $remote_server])
+shopt -s expand_aliases
+source ~/.bashrc
 echo "Caching configs...";
-php73 artisan config:cache
+php73 {{ $source_dir }}/ {{ $release_dir }}/{{ $release }}/artisan config:cache
```
chrispelzer: This has a space in between the {{ $source_dir }}/ and {{ $release_dir }}

Envoy.blade.php
```diff
@@ -264,11 +264,13 @@
...
 echo "Caching routes...";
-php73 artisan route:cache
-echo "Done caching";
+php73 {{ $source_dir }}/ {{ $release_dir }}/{{ $release }}/artisan route:cache
```
chrispelzer: Same thing here

Envoy.blade.php
robertvrabel: Thanks for noticing! I copied it from https://github.com/waynestate/base-site/blob/cfe9ac46b6a64217ea8a5da4fad7a3a1ff3f5159/Envoy.blade.php#L260

Seems like it works still, but I will fix all instances of the extra space

nickdenardis:
👍

PR #388, opened 2019-08-29
Feature/individual views styleguide

robertvrabel:
* Move the addthis config as a script tag that is appended to the body. This was the only way to get the config to actually work (it was never doing anything prior to this PR). This fixes the tabindex accessibility issues with addthis.
* Bring the individual news and profile views into the styleguide menu.json so they are checked for accessibility issues.
* Add exceptions for blockquotes now that the individual news view is being checked.


breakdancingcat:
Thank you!

PR #401, opened 2019-09-11
hotfix/6.11.1

robertvrabel:
The new ignition error page requires that the `config/auth.php file is present with certain values (since it hinges on this to check what to report to you). I've added the minimal needed configuration in order to get the error page to properly work.

### DIFF
--- a/config/auth.php
+++ b/config/auth.php
@@ -1,0 +1,68 @@
+<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Authentication Defaults
+    |--------------------------------------------------------------------------
+    |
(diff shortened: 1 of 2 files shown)


config/auth.php
```diff
@@ -0,0 +1,68 @@
...
+    | be assigned to any extra authentication guards you have defined.
+    |
+    | Supported: "database", "eloquent"
+    |
+    */
+
+    'providers' => [
+        'users' => [
+            'driver' => 'eloquent',
+            'model' => null,
```
robertvrabel: The default is `App\User` but since we don't have a user model I set it to null instead.

PR #412, opened 2019-10-10
Changing <p> to <li>

Fw7424:
Currently "No profiles found" in list, but not a list item. Axe sees this as an error with the list structure. Changing to `<li>` ensure this is read as being in the list.


chrispelzer:
Make sure you update the `package.json` with the latest version

PR #415, opened 2019-11-01
Accessibility reports flagged this as a redundant link

breakdancingcat:
The site title in the header and the footer were both named the same thing and linking to the same place


nickdenardis:
I'm OK with not allowing a link in the footer by default and if one is needed to add it to that specific side codebase.

If this gets merged in, we'll likely want to remove the "Link" field from the default Footer promotion group to avoid confusion since it won't be used.

robertvrabel:
One reason this feature exists currently is when you are on a subsite, the main + subsite footers are represented and the link is somewhat useful to browse back to the main site. https://base.wayne.edu/styleguide/footer/contact/two . We sometimes use the "surtitle" in the config to also accomplish this in the header - but it isn't always the case since it doesn't do it by default.

Is there a way we could leave the link but describe it differently for accessibility?

nickdenardis:
Feels like the content is causing the accessibility conflict which is separate from the flexibility of the implementation. 

Can we add some description to the promotion group in the CMS to specify intended use for the "Link" field?

Once we get to a Two+ footer columns the "Title" and "Link" fields are ignored in the CMS and it is freeform content only. 

Removing this "Link" will still result in content causing accessibility conflicts. 

Perhaps the wording of "Link - used only to link to a destination other than the homepage of the website" or something like that to denote there is no need to link to the homepage since the header of the page already links to the homepage by default.

breakdancingcat:
Wouldn't that result in the site title and footer title having the same text going to _different_ destinations?  

If we're solving for: 
**The footer title link is useful when on a subsite, with combined parent/sub footers, no surtitle.** 

1. If the department doesn't want to use the surtitle, do they see a need to link to the parent site?

2. The College of Engineering and The College of Education have entire headers dedicated to the parent site title. Law School opted out of the subsite title appearing in the header at all, and the department/subsite name is only in the footer (and the title of the page). The prioritization of how a college's hierarchy should be displayed is taken into account on a case by case basis already. 

3. If we remove the footer title link, the user has the opportunity to be intentional with their footer content whether it's supplying contact information directly or linking to a contact/about/staff page if it's not in the top nav. 

I'm still on board with not forcing a link for the title of the column(s).

nickdenardis:
@breakdancingcat 
I see your point. I'm good with cutting this link by default and adding it in _if_ needed on a per-site basis.

PR #420, opened 2019-11-21
Feature/profile listing image ratio portrait by default

tomkrupka:
Updated profile listing page grid images to be portrait ratio by default

### Before:
Screen Shot 2019-11-21 at 3 04 08 PM

### After:
image


breakdancingcat:
Thanks! 🔥

PR #430, opened 2020-01-27
Main Under Menu Merge with Subsite

chrispelzer:
Similar to how the Merge with Main Contact functions but for the Under Menu. It seems site have been needing this more and more.

### DIFF
--- a/app/Repositories/PromoRepository.php
+++ b/app/Repositories/PromoRepository.php
@@ -131,6 +131,14 @@ class PromoRepository implements RequestDataRepositoryContract, PromoRepositoryC
             $promos['contact'] = $main_contact;
         }
 
+        // Inject the main under menu if we are on a subsite
+        $main_under_menu = $promos['main_under_menu'] ?? [];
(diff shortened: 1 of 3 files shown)


nickdenardis:
Where is the previous feature documented and where should we put the documentation for this so others know it exists?

Looking at the options, perhaps if we will expand this beyond one area we could use a more generic key in the config, something like `merge_with_parent` instead of `merge_with_main_contact` and `merge_with_main_under_menu`.

chrispelzer:
It's the same as the merge_with_main_contact at https://github.com/waynestate/base-site/blob/develop/config/base.php#L191-L262

I'm not sure about the documentation. It is just part of the site config.
The only other part that talks about it is at https://github.com/waynestate/base-site/blob/master/README.md#configure-the-site

breakdancingcat:
@nickdenardis @chrispelzer I would like to have configuration information for different parts of Base. In some cases I've put the info in the styleguide page. For under menu and footer contact, we can put the merge information on those styleguide pages. Thoughts?

chrispelzer:
@breakdancingcat Styleguide would work, though the Styleguide has been primary for the viewing of the site and not so much the configuration, but I think it's probably the best place since it could also be a living document for the configuration of custom stuff on the site as well.

PR #432, opened 2020-02-04
Pre Push Hook

traceremick:
Added pre push hook to catch version mismatch (lack of version bump) before merging into live code.

The following error occurs when the package.json version is different from the branch version:

Screen Shot 2020-02-04 at 2 16 22 PM


breakdancingcat:
Thank you @traceremick !

PR #433, opened 2020-02-06
Rate Limiting

traceremick:
Adding page throttling for excessive page load:

Screen Shot 2020-02-06 at 11 09 35 AM


traceremick:
Added trust proxy which overrides the default ip with the forwarded ip address (for load balancers)

PR #436, opened 2020-02-11
Feature/profile groups

Fw7424:
Refactor `getProfilesByGroup` to allow for profiles to exist in multiple groups.


app/Repositories/ProfileRepository.php
```diff
@@ -80,19 +80,12 @@ public function getProfilesByGroup($site_id)
...
-            });
-        })
-        ->keyBy('AccessID')
-        ->groupBy('group', true)
+        $grouped = collect($all_profiles['profiles'])->keyBy('data.AccessID')
+        ->groupBy([
+            function ($profile) {
+                return $profile['groups'];
+            },
+        ], $preserveKeys = true)
```
nickdenardis: We talked about changing this away from the assignment format and instead pass the boolean:

`], true) // Preserve Keys`

PR #440, opened 2020-02-20
Profile and directory adjustments

breakdancingcat:
* Corrected the SVG drop down arrow on the profile listing page
* Changed the profile listing and directory templates to display 4 profiles per row instead of three
    * Individuals have a hard time providing quality images for that space
    * More users per line, appears more like a yearbook
* Directory template uses portrait size instead of square
* Reconfigured the grid template
    * Each element is wrapped in an a tag if a link is present 
    * Only the title is green and underlined if it is a link, instead of all the text being green
* Extended inset in tailwind.js, ex: right-4 = 1rem from the right

|Before|After|
|-----|-----|
|directory-og|directory|
|profile-og|profile-listing|
|grid-og|grid|

### DIFF
--- a/resources/scss/components/_formy.scss
+++ b/resources/scss/components/_formy.scss
@@ -54,7 +54,7 @@
     }
 
     select {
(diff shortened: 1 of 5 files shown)


Fw7424:
:moneybag:

PR #447, opened 2020-06-08
Clean up the logic in the grid repository

chrispelzer:
Rely on the inverse of the if statement since we are requiring the page field, if we don't have it, we can't move on.


Fw7424:
I wonder if we should consider setting this up to allow multiple grid promos by default.

PR #448, opened 2020-06-09
Hotfix/6.15.1

Fw7424:
Visually no change. This proposed solution will resolve the issue for adding accordions within the content. 
 [TP3 ticket:](https://waynedotedu.tpondemand.com/entity/100940-base-update-accordion)

### DIFF
--- a/resources/scss/components/_accordion.scss
+++ b/resources/scss/components/_accordion.scss
@@ -19,8 +19,7 @@
                 @apply .bg-grey-darker .text-white;
             }
 
-            // Open symbol
-            &.symbol::before {
+            &::before {
                 @apply .absolute .right-0 .pr-4 .top-0 .pt-4;
 
                 content: '+';
@@ -28,7 +27,7 @@
         }
 
         // Close symbol
-        &.open > .symbol::before {
+        &.open > a::before {
             content: '\2013';
         }
 
--- a/resources/scss/components/_accordion.scss
+++ b/resources/scss/components/_accordion.scss
@@ -20,7 +20,7 @@
             }
 
             // Open symbol
-            & span::before {
+            &.symbol::before {
                 @apply .absolute .right-0 .pr-4 .top-0 .pt-4;
 
                 content: '+';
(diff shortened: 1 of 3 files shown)


InnocentSmasher:
Since every accordion will need the symbol, is there any way to bind it directly to the <a> without needing the class? It might create a bit more specificity (.accordion > li > a::before) but it would be more robust and leaves less room for user error.

breakdancingcat:
@Fw7424  
I know it doesn't auto populate, but can you use the PR template as well? Ty :)

breakdancingcat:
> Since every accordion will need the symbol, is there any way to bind it directly to the without needing the class? It might create a bit more specificity (.accordion > li > a::before) but it would be more robust and leaves less room for user error.

I like this solution, `&::before` &::after` instead of needing the class. 
Can you try this and also test that this won't affect content by making a list with links inside of the accordion content? Thank you!

Fw7424:
@breakdancingcat I tested with a link inside the accordion. It doesn't seem to affect it. although i'm not sure what other weird stuff we might run into, in the future by removing the specificity of using a class.

chrispelzer:
Need a version bump as well on this.

nickdenardis:
Is this working on an example site? 

It will need a full manual accessibility test with Voiceover and NVDA before it is merged into Base.

breakdancingcat:
@nickdenardis There's no change to anything accessible in this PR, however I did just notice that when the user closes the option, aria-expanded doesn't change to false until another one opens. Not sure that's a deal breaker, but this PR is good.

breakdancingcat:
So far I found that screen readers will say "plus" now.
I found we can add alternative text for pseudo elements, but safari doesn't like it and won't render the element. 
```
 content: "+" / "";
```
https://www.w3.org/TR/css-content-3/#accessibility
Still looking into this

nickdenardis:
@breakdancingcat 
This is still assigned for a full accessibility review before being merged.

breakdancingcat:
Declining this one because of the plus being read, I will change the span to a div and leave it as it was

breakdancingcat:
Declining this one because of the plus being read, I will change the span to a div and leave it as it was

PR #454, opened 2020-07-30
A11y fixes for off-canvas

nickdenardis:
Three accessibility improvements.

1. When the off-canvas menu is open, the "Menu" button is renamed to "Close" since the keyboard is trapped in the menu until the user closes or hits "Esc"
1. The "Esc" key now returns the user back to the menu button that expanded the off-canvas menu instead of being hidden off-screen
1. For sites that have a top menu and a local page menu the "Main menu" in the off-canvas can be confusing, this moves it to a config value so it can be changed based on the context, ex. "CLAS menu"

| Before | After |
|--------|--------|
| Screen Shot 2020-07-30 at 4 02 37 PM | Screen Shot 2020-07-30 at 4 02 14 PM |

### DIFF
--- a/resources/views/components/content-area.blade.php
+++ b/resources/views/components/content-area.blade.php
@@ -16,7 +16,7 @@
                     <div class="slideout-main-menu mt:hidden">
                         <ul class="main-menu mb-2">
                             <li>
-                                <a role="button" class="main-menu-toggle pt-2 pb-2 pl-3 pr-3 block" tabindex="0" aria-expanded="false">Main Menu</a>
+                                <a role="button" class="main-menu-toggle pt-2 pb-2 pl-3 pr-3 block" tabindex="0" aria-expanded="false">{{ config('base.top_menu_label') }}</a>
 
                                 {!! $top_menu_output !!}
                             </li>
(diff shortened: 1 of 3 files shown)


resources/views/components/content-area.blade.php
```diff
@@ -16,7 +16,7 @@
                     <div class="slideout-main-menu mt:hidden">
                         <ul class="main-menu mb-2">
                             <li>
-                                <a role="button" class="main-menu-toggle pt-2 pb-2 pl-3 pr-3 block" tabindex="0" aria-expanded="false">Main Menu</a>
+                                <a role="button" class="main-menu-toggle pt-2 pb-2 pl-3 pr-3 block" tabindex="0" aria-expanded="false">{{ config('base.top_menu_label') }}</a>
```
breakdancingcat, line 19: You could use short site tile or base config here if interested!

nickdenardis: Good point. I've found a few in the CMS that is not actually short and a replica of the full site title. Ideally, this would be less than 16 chars. If that limit existed in the CMS I'd be totally comfortable.

Also, the dealbreaker to using short title field is that this label is meant to be displayed on the subsites (which wouldn't have an accurate short title for this context).

Let's leave it in code for the moment and have the discussion if it becomes an issue.

PR #455, opened 2020-07-31
Feature/revert blockquote

breakdancingcat:
## Reason for change
* Concluded we would need additional elements or images beyond before/after pseudo elements to make a complete fancy quote style
* Fancy quote styles wil be additional beyond what's default in content, either designed and available through the components area of the styleguide in the future or designed per site 
* Reverting back to a web-default blockquote approach, but included cite tags as mentioned in a previous PR 

Before | After
-- | --
blockquote-before|blockquote-after

### DIFF
--- a/resources/scss/components/_content.scss
+++ b/resources/scss/components/_content.scss
@@ -29,37 +29,15 @@
     }
 
     blockquote {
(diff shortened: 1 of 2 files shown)


Fw7424:
LGTM

PR #456, opened 2020-07-31
Adding default reviewers

breakdancingcat:
## Reason for change
Adding default reviewers

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @waynestate/wayne-state-web-team


traceremick:
@breakdancingcat just noticed on success site PR: do we want this file to be included on every base site?

https://bitbucket.org/waynestate/success/pull-requests/37/applying-patch?link_source=email#LCODEOWNERST1

breakdancingcat:
@traceremick ah heck, it's not a deal breaker but if it's only on the base repo that would be sufficient.

chrispelzer:
@breakdancingcat @traceremick it doesn't matter as long as it isn't causing issues. We carry along the `.travis.yml` to all projects and it's used only for Base PRs to run Travis CI

PR #464, opened 2020-09-25
Error handler was erroring...

chrispelzer:
`getStatusCode()` does not exist in php core Exception but within the Symfony HttpException, since we only have custom views for HTTP error codes we can rely on making sure the exception is an instance of HttpException otherwise give it a standard 500 error view


nickdenardis (approved):
👍

PR #465, opened 2020-09-28
adding tailwind float classes to the float:none

InnocentSmasher:
Since we're intentionally setting "float:none" when the user inlines "style=float:none", we should also be overriding the float-right and float-left classes from tailwind.


InnocentSmasher:
Old:
base-old

New:
base-new

nickdenardis (commented):
:+1:

breakdancingcat:
Thank you!

PR #466, opened 2020-09-29
skipping surtitle in menu-top when surtitle_disabled is true

InnocentSmasher:
If a subsite has the property surtitle_disabled set to true, the surtitle will not show in menu-top.blade.php.


nickdenardis (commented):
👍

PR #471, opened 2020-10-28
Feature/directory order by group

InnocentSmasher:
Giving the users the ability to display only certain groups on a directory page, as well as choose the order those groups are displayed in. Works non-invasively in the standard directory template (as in, if you don't supply the appropriate custom fields, it defaults to the normal/existing template).

directory-order-pipe

### DIFF
--- a/app/Repositories/ProfileRepository.php
+++ b/app/Repositories/ProfileRepository.php
@@ -124,6 +124,8 @@ class ProfileRepository implements ProfileRepositoryContract
 
         $group_order = explode('|', $groups);
 
+        $profiles = [];
+
         // Retain the order of the groups as they were piped in
         foreach ($group_order as $group) {
             foreach ($profile_listing['profiles'] as $profile) {
--- a/app/Repositories/ProfileRepository.php
+++ b/app/Repositories/ProfileRepository.php
@@ -118,6 +118,27 @@ class ProfileRepository implements ProfileRepositoryContract
     /**
      * {@inheritdoc}
      */
+    public function getProfilesByGroupOrderPiped($site_id, $groups)
+    {
+        $profile_listing = $this->getProfiles($site_id);
+
+        $group_order = explode('|', $groups);
+
+        // Retain the order of the groups as they were piped in
+        foreach ($group_order as $group) {
+            foreach ($profile_listing['profiles'] as $profile) {
+                if (array_key_exists($group, $profile['groups'])) {
+                    $profiles['profiles'][$profile['groups'][$group]][] = $profile;
+                }
+            }
+        }
+
+        return $profiles;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function getDropdownOptions($selected_group = null, $forced_profile_group_id = null)
     {
         // Default Options
--- a/tests/Unit/Repositories/ProfileRepositoryTest.php
+++ b/tests/Unit/Repositories/ProfileRepositoryTest.php
@@ -266,9 +266,7 @@ class ProfileRepositoryTest extends TestCase
 
         $groups = collect($return_user_listing)->map(function ($item) {
             return array_shift($item['groups']);
-        })->unique()->toArray();
-
-        krsort($groups);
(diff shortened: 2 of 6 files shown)


app/Repositories/ProfileRepository.php
```diff
@@ -115,6 +115,29 @@ public function sortGroupsByDisplayOrder($grouped, $groups)
...
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function getProfilesByGroupOrderPiped($site_id, $groups)
+    {
+        $profile_listing = $this->getProfiles($site_id);
+
+        $group_order = explode('|', $groups);
```
traceremick, line 125: Why a pipe instead of a comma?

app/Repositories/ProfileRepository.php
```diff
@@ -115,6 +115,29 @@ public function sortGroupsByDisplayOrder($grouped, $groups)
             ->toArray();
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function getProfilesByGroupOrderPiped($site_id, $groups)
```
traceremick, line 121: Think "piped" should be removed or renamed.

tests/Unit/Repositories/ProfileRepositoryTest.php
```diff
@@ -254,4 +254,38 @@ public function profiles_should_be_grouped()
...
+    public function profile_group_ids_should_return_ordered_array()
+    {
+        // Mock the user listing
+        $return_user_listing = app('Factories\Profile')->create(10);
+
+        $groups = collect($return_user_listing)->map(function ($item) {
+            return array_shift($item['groups']);
+        })->unique()->toArray();
+
+        krsort($groups);
```
traceremick, line 271: This can be done in the collection above?

tests/Unit/Repositories/ProfileRepositoryTest.php
```diff
@@ -254,4 +254,38 @@ public function profiles_should_be_grouped()
...
+
+        $piped_groups = implode('|', array_keys($groups));
```
traceremick, line 273: This line is interesting. I don't quite understand it.

app/Repositories/ProfileRepository.php
InnocentSmasher: This is being pulled in from custom fields, and the "standard" we've been using has been a pipe. It's done on the Law site (where the code came from) and there have been a few sites since that use a pipe to denote different profile_group_ids.

tests/Unit/Repositories/ProfileRepositoryTest.php
InnocentSmasher: Chris and I spent over two hours getting to a "working" state. We know there's cleanup. The krsort here is because we went back and forth between collections and arrays a few times and were getting stuck, so we just left this as a "hey at least it works" state.

tests/Unit/Repositories/ProfileRepositoryTest.php
InnocentSmasher: Since the whole point of the "getProfilesByGroupOrderPiped" is to take piped in group ids, we have to turn the available group ids from the faker data into the same format it'll be working with: '0|1|2|3'

nickdenardis (approved):
Although there are some code cleanup opportunities here, this looks good!

app/Repositories/ProfileRepository.php
InnocentSmasher: Does anyone have any suggestions for a better name? I also think it's wordy, but dropping "Piped" makes it sound too similar to both "getProfilesByGroup" and "sortGroupsByDisplayOrder". There honestly is probably a better way to name this, but I feel like I've been looking at it for too long and I'm too close to it to come up with an effective one.

traceremick: If "getProfilesByGroupOrder" is what the function does then thats an OK name, regardless of how close it is to others.

PR #473, opened 2020-11-09
Transition from Google Analytics JS to Google Tag Manager

nickdenardis:
## Reason for change

Prepping the move to GTM since the Google Analytics v4 requires a completely different JS interaction and handles a lot of the click, scroll, and movement tracking that we have done by hand before.

We don't have all the properties in GTM yet, there are hundreds and still deciding on the best way to manage them. We may have to have a fallback to the regular GA code.

Working to implement this slowly on new sites before doing any large scale conversion.


breakdancingcat:
@nickdenardis any update on this one?

breakdancingcat (approved):

nickdenardis:
@breakdancingcat So this is the method to convert an existing site to GTM from GA, we are not yet ready to have all new sites using GTM but are slowly moving over to it. 

I should refactor this to be conditional-based so we could use GA or GTM depending on the config so we don't have a longstanding PR out there. 

I'll get that done and re-ping everyone on the upate.

PR #474, opened 2020-11-09
Feature/contact table template

InnocentSmasher:
Giving users ability to set up a contact table (with or without a table of contents).
screencapture-base-wayne-local-styleguide-contacttables-2020-11-09-11_08_40


resources/views/contact-tables.blade.php
```diff
@@ -0,0 +1,53 @@
...
+                        <th class="w-48">Name</th>
+                        <th>Title</th>
+                        <th class="w-40">Office</th>
+                        <th class="w-40">Phone</th>
+                    </tr>
+                    </thead>
+                    <tbody>
+                    @foreach($profile_list as $profile)
+                        <tr>
+                            <td>@if(isset($profile['data']['Email']))<a href="mailto:{{$profile['data']['Email']}}">@endif{{$profile['data']['First Name']}} {{$profile['data']['Last Name']}}@if(isset($profile['data']['Email']))</a>@endif</td>
```
nickdenardis, line 39: For all these, do we need these conditionals here? Pretty sure, although needs confirmation, if `{{$profile['data']['Title']}}` is in Blade and there is no key in that spot it will display nothing instead of erroring.

resources/views/contact-tables.blade.php
```diff
@@ -0,0 +1,53 @@
+@extends('components.content-area')
+
+@section('content')
+    @include('components.page-title', ['title' => $page['title']])
+
+    <div class="content">
+        {!! $page['content']['main'] !!}
+    </div>
+
+    <a id="directory-contents"></a>
```
nickdenardis, line 10: Can you check to see if this errors from an automated accessibility check? This tag with no contents likely isn't something we could put live.

styleguide/Pages/ContactTablesNoTOC.php
```diff
@@ -0,0 +1,25 @@
...
+        return app('Factories\Page')->create(1, true, [
+            'data' => [
+                'profile_group_id' => '0|1|2|3',
+                'table_of_contents' => 'hide',
+            ],
+            'page' => [
+                'controller' => 'ContactTableController',
+                'title' => 'Contact tables, no table of contents',
+                'id' => 101114,
+                'content' => [],
```
nickdenardis, line 21: Can the page content for this styleguide page explain how to use it to anyone viewing the styleguide instead of requiring a developer to dive into the code to add the configuration to the CMS to use it.

nickdenardis (commented):
Here is my feedback. A number of small things and questions.

PR #480, opened 2020-12-18
Hotfix/6.18.1

chrispelzer:
Add some context behind the redis:clear command for how many keys were removed when running `redis:clear`


nickdenardis (commented):
👍

PR #488, opened 2021-02-01
Hotfix/6.20.1

InnocentSmasher:
Reverting decision to hide breadcrumbs on print

print-breadcrumbs-update

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "base",
   "private": true,
-  "version": "6.20.0",
+  "version": "6.20.1",
   "description": "",
   "scripts": {
     "dev": "npm run development",
--- a/resources/scss/components/_print.scss
+++ b/resources/scss/components/_print.scss
@@ -18,6 +18,10 @@
         content: '';
     }
 
+    #breadcrumbs-menu a[href]::after {
+        content: '';
+    }
+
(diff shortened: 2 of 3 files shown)


InnocentSmasher:
I'm not crazy about the href showing on the breadcrumbs. I'm fine with it either way. However, hiding the href isn't feasible with the way it's set up.

If we want to hide it, what is the opinion on adding a wrapper class to the breadcrumbs component so it is targetable from the SCSS? There isn't a great way to target it otherwise (involves a very long string of selectors that would also be very fragile).

@nickdenardis

nickdenardis:
The breadcrumbs add a lot of value to the printed version of the page but you're right, the URLs should not be shown.

The other two `<nav>` elements have `id` attributes:
https://github.com/waynestate/base-site/blob/master/resources/views/components/content-area.blade.php#L14
https://github.com/waynestate/base-site/blob/master/resources/views/components/menu-top.blade.php#L46

Let's do that here, `id=breadcrumbs-menu` to target these links in a non-brittle way.

InnocentSmasher:
Updated to remove links after breadcrumbs.

Screen Shot 2021-02-02 at 9 18 04 AM

nickdenardis (commented):
Looks good to me!

PR #489, opened 2021-02-02
fixing stylelint issue (oops)

InnocentSmasher:
noticed a small hierarchy error with stylelint. not really sure why it's an issue in this case (i understand the purpose of the error), but now it's fixed.


nickdenardis (commented):
👍

PR #490, opened 2021-02-02
Feature/accordion plus minus update

InnocentSmasher:
Auto-appending `<span aria-hidden="true"></span>` to accordion buttons that don't already have it.

Before | After
--- | ---
accordions-before | accordions-after

### DIFF
--- a/resources/js/modules/accordion.js
+++ b/resources/js/modules/accordion.js
@@ -45,6 +45,11 @@ import 'accordion/src/accordion.js';
         item.querySelectorAll('li a:first-child').forEach(function(item) {
             item.setAttribute('role', 'button');
             item.setAttribute('aria-expanded', 'false');
+
+            // Adds +/- span if there isn't one already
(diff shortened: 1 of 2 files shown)


InnocentSmasher:
Goal was to keep this as "simple" as possible. Since <span> isn't allowed in our CMS (unless you hit publish in the source view), this checks if each item has a span. If it doesn't, it inserts the span markup just inside the accordion button (before any title/text content) to match the component placement.

I messed around with it in VoiceOver and the accordion in the content behaved the exact same way that our promo accordions do.

After this gets an approval, I will bring it into develop so Krystal/Nick can verify that it's all good.

traceremick:
Very nice.

traceremick (approved):
Very nice

nickdenardis (commented):
👍 Nice! Assign Krystal and me a ticket to confirm it for accessibility.

PR #494, opened 2021-02-24
Hotfix/6.20.5

InnocentSmasher:
The `@responsive` style inside the `@screen print` media query was escaping that print scope for some reason. Since `@responsive` isn't a normal CSS thing, it was causing specificity issues.


nickdenardis (commented):
👍

PR #495, opened 2021-02-25
Hotfix/6.20.6

breakdancingcat:
Nested unordered and ordered lists don't need a bottom margin

|Before|After|
-------|-------
|Screen Shot 2021-02-25 at 11 06 33 AM|Screen Shot 2021-02-25 at 11 06 41 AM|


InnocentSmasher (approved):

traceremick:
Does this cause issues with the visual impaired?

nickdenardis:
Feel like there should be some spacing, perhaps the same amount as between paragraphs?

Do we have an example of these nested UL's outside of a table of contents?

InnocentSmasher:
(Personal opinion) I don't think you need any extra spacing besides what is provided with the built in line-height of each `<li>`. I opened up a Word document just to see what the default behavior for nested lists was, and there is no additional space. Google Docs is the same way.

To me, the extra space provides a harsh disconnect between the overall list structure.

Rolaine:
mmmm…. Sorry to intervene. I have to say something. The before and after is great, helps me see what you guys are talking about! Has that always been the case? Anyways, I prefer having space. Having the text underline make the line height soo tight when in default. If there weren’t any underlines, maybe default spacing is ok. I always add spacing in our mocks and it makes reading a lot more pleasurable and easier on the eyes.

breakdancingcat:
> Does this cause issues with the visual impaired?

@traceremick  No, I believe it was an error in our styles to begin with. As long as the line height is appropriate, and it is, I can't find why it would cause issue.

breakdancingcat:
I haven't found anything that points to lists needing space after the nested ul. I believe this was a specificity error in our CSS to begin with. Like Chuck said, the only thing I found regarding accessibility is line height, which ours is appropriate. I don't think the space helps, I think it causes a break in continuity.

Rolaine:
Ok ok. I’ll go with out bottom margin :D I tested it on TP3 comment. I don’t do Word (sorry Chuck). Word is never pretty. But can we bump up the line height a bit to make up for the tightness? Or is that optical illusion and both have the same line height?

breakdancingcat:
This is the first example of nested lists I could think of, if there were large breaks after each UL, the following items would feel disconnected. https://policies.wayne.edu/academics/14-1-policy-on-course-schedule-types--credit-hours-and-delivery-modes

This is the closest thing I can find to "accessible nested lists" https://www.w3.org/WAI/tutorials/page-structure/content/

breakdancingcat:
The list line height is the same as paragraph line height, here is an increased line height. I'm all for creating space, but I'm curious what the thought behind a list needing increased line height vs a paragraph. 

image

nickdenardis:
I was only commenting on the spacing between lists (or list items), independent of the line-height within a list item. We have traditionally kept the individual list item line-height tight so when a list item is long enough to wrap multiple lines it is clear what all is contained in them and the space between them (or lists) is where the spacing would be. 

Looking at the screenshots the spacing on the left seems most appropriate as a default and we can make case-by-case decisions to expand it on sites if needed.

PR #509, opened 2021-06-08
Hotfix/6.22.1

Fw7424:
Added default for meta tags. 
[116259 Base - Add default meta images](https://waynedotedu.tpondemand.com/entity/116259-base-add-default-meta-images)


nickdenardis (approved):
👍

PR #516, opened 2021-08-06
Feature/grid spotlight combo

breakdancingcat:
* Condense the base menu, templates that are promo driven are in their own submenu, all profile data driven templates are in their own submenu, same with news
* Remove spotlight language, be able to use custom fields to turn on an individual promo item view or use the promotion item's link (this will be documented) 
image


nickdenardis (approved):
👍

PR #520, opened 2021-08-18
Moved new Base features to a submenu in the styleguide 

nickdenardis:
## Background

When adding a new feature to Base the menu item is appended to the top-level "Templates" menu item and the `menu_item_id` is incremented by 1. This is logical and doesn't cause any issue until you try to update a site to the newest version of Base. 

When updating a site, the menu.json from the newest Base will try to merge with the site-specific styleguide menu, if any new templates were added to Base that conflict with the incrementing `menu_item_id` items from the site-specific features it will cause a conflict and force the developer to hand update the `menu.json` file and potentially re-number all the site-specific menu items.

## Proposal 

Since we have now broken down the "Templates" top-level menu item into sub-items based on the type of template, this change. uses that same approach to consolidate all site-specific features/templates into their own top-level menu item.

## Result

This will allow all site-specific features to start with a unique number and in a spot that doesn't conflict with potentially incoming menu items.

## Related

Updated all the numerical values in the `menu.json` file to be digits instead of strings to match consistency between each other and the new features being generated.

## Demo

| Before | After |
|--------|--------|
| Screen Shot 2021-08-18 at 11 34 56 AM | Screen Shot 2021-08-18 at 11 33 49 AM |

## Landing page

The landing page for the site-specific features includes the instructions from the README on how to create one.

Screen Shot 2021-08-18 at 11 38 16 AM

### DIFF
--- a/app/Console/Commands/BaseFeature.php
+++ b/app/Console/Commands/BaseFeature.php
@@ -82,14 +82,14 @@ class BaseFeature extends Command
     {
         $menu = $this->getMenu();
 
-        $item = end($menu[101]['submenu']);
+        $item = end($menu[101]['submenu'][999]['submenu']);
 
         $item['menu_item_id']++;
         $item['page_id'] = $item['menu_item_id'];
         $item['display_name'] = $this->feature.'s';
         $item['relative_url'] = '/styleguide/'.strtolower($this->feature).'s';
 
-        $menu[101]['submenu'][$item['menu_item_id']] = $item;
+        $menu[101]['submenu'][999]['submenu'][$item['menu_item_id']] = $item;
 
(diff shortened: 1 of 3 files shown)


traceremick (approved):
Is this backwards compatible with diff base upgrades?

tomkrupka:
LGTM

tomkrupka (approved):

nickdenardis:
> Is this backwards compatible with diff base upgrades?

Tried the upgrade with two types of sites, one that is pretty recent and one that was older.

The older site, from 6.15.0, worked the same as typical, there were a few conflicts but mostly they had to do with the nesting of styleguide menu items from #516 

For a semi-recent site, the only menu diff is the new submenu added:

Screen Shot 2021-08-19 at 7 03 42 AM

breakdancingcat (approved):

PR #522, opened 2021-09-27
Feature - #123009 Base - Content.scss

AmbyrElan:
## Reason for change

Visually group flow of text after headers

## Required checklist

- [x] Provide TP3 link: [#123009 Base - Content.scss](https://waynedotedu.tpondemand.com/restui/board.aspx?#page=userstory/123009)
- [x] Screen shot or video of before/after - Screenshots will be added after

## Reminders

- SiteImprove accessibility check
- Verified each page looks good on each viewport size
- Add an example to the styleguide if applicable
- Add tests to prove the code works
- Provide CMS site link if applicable

## Screenshots / Videos

| Before | After |
|--------|--------|
old_base-wayne-edu-styleguide-profiles-2021-09-28-11_00_00|new_base-wayne-local-styleguide-profile-aa0000-2021-09-28-11_10_10

### DIFF
--- a/resources/scss/components/_global.scss
+++ b/resources/scss/components/_global.scss
@@ -46,22 +46,32 @@ h1 {
 
 h2 {
     @apply .text-3xl;
+
+    margin-bottom: 0.5rem;
 }
 
 h3 {
     @apply .text-2xl;
+
+    margin-bottom: 0.3rem;
 }
 
 h4 {
     @apply .text-xl;
+
+    margin-bottom: 0.3rem;
 }
 
 h5 {
     @apply .text-lg;
+
+    margin-bottom: 0.3rem;
 }
 
 h6 {
     @apply .text-sm;
+
+    margin-bottom: 0.3rem;
 }
 
 // Firefox uses the outline focus color as the text color. Since our overall background is white we need to change it.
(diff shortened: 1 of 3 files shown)


resources/scss/components/_global.scss
```diff
@@ -46,22 +46,32 @@ h1 {
 
 h2 {
     @apply .text-3xl;
+
+    margin-bottom: 0.5rem;
```
nickdenardis, line 50: Is there an existing tailwind class that is this same spacing we could utilize here in an `@apply` instead of a direct number in the CSS?

nickdenardis (commented):
One inline comment.

breakdancingcat:
@nickdenardis the description says screenshots will be added. We were having issues with the code formatter so this isn't done yet.

resources/scss/components/_global.scss
breakdancingcat: I had her type these in explicitly. We will convert them.

breakdancingcat: @AmbyrElan can you add ml-2 to line 48 and all the other headings,
`@apply .text-3xl .ml-2;` 
And for h3 through h6 
Add ml-1 in the same manner. 
Sorry about that!

breakdancingcat:
Updated these on my latest hotfix

PR #524, opened 2021-09-28
hotfix/6.23.1 - Promo view was scoped to the wrong variable from an older iteration

breakdancingcat:
Promo view was scoped to the wrong variable from an older iteration
Also we don't have styling for pre tag at this time so changed pre tag to code tag on the site specific instructions


nickdenardis (commented):
👍

PR #527, opened 2021-12-03
Base 7.0

breakdancingcat:
# Introducing Base 7.0 🎉
- Upgrading to Tailwind 2.0
- Colors now use numbers 50-900 instead of darkest-lightest
- Renamed the WSU #ffcc33 yellow to gold
- Renamed the Styleguide "Content area" menu to "CMS Content examples" and separated the CMS styles into their own pages 
    - The new "Content area" menu item shows basic body copy examples as they appear from the WYSIWYG
    - Added fourish heading styles available to end users and developers
    - Created new button classes that allow more flexibility from end users in the CMS and developers when using them
    - Modified the video component and turned it into an easily usable class in both the WYSIWYG and templates; You can apply the play button with a simple class on the parent
    - Added all the tables to one page and allow them to each scroll horizontally on mobile devices instead of widening the whole page/content area and added more directions
    - Updated the figure page with current classes
    - Added commonly requested or used basic layouts for end users to paste into the WYSIWYG
    - Used Tailwinds new color display and applied our colors
- Separated Components from main layout pieces
    - Removed header w/o short title because short title is a requirement from the CMS and not optional
    - Additional footer columns now have headers, a little smaller than the first column
- Updated the under menu and button components and their instructions to be clear to the end users
- Added tiktok to the social footer and made it a bit smaller
- Darkened the footer and social bar colors


chrispelzer (approved):

nickdenardis:
👍

Fw7424:
👍

PR #528, opened 2021-12-10
Update to PHP8

bannakaffalata:
Update to PHP8 and fix tests and areas that need updating based on a more strict syntax within PHP8


nickdenardis (commented):
👍

PR #530, opened 2021-12-14
Feature/base 7 tweaks already

breakdancingcat:
- Added gold button option and consolidated button code
- Added more instructions to the under menu buttons
- "All topics" option now in topics menu on the news article listing page 
- Events can appear in a whole row
- Remove duplication of top menu when forcing the left menu to show 
- Adding news to places where we reference articles for context


nickdenardis (commented):
👍

app/Repositories/TopicRepository.php
```diff
@@ -98,7 +105,11 @@ public function sortByLetter($topics)
     public function setSelected($topics, $topic)
     {
         return collect($topics)->map(function ($item) use ($topic) {
-            $item['selected'] = $item['slug'] === $topic ? true : false;
+            if (!empty($topic)) {
+                $item['selected'] = $item['slug'] === $topic ? true : false;
```
traceremick, line 109: these can be returned instead of doing a ternary:

$item['selected'] = $item['slug'] === $topic;

$item['selected'] = $item['slug'] === '';

breakdancingcat: Awesome, I'll work on this. Thanks!

PR #534, opened 2022-01-04
hotfix/7.0.1 - Scrollbars were automatically showing on content area for some users

breakdancingcat:
Fixing scrollbars automatically showing on content area for some users

|before|after|
|------|-----|
|basebefore|baseafter|


nickdenardis (commented):
👍

PR #541, opened 2022-01-18
Update styleguide.blade.php

krystosch:
Added a little more context to headings and fixed a typo.


nickdenardis (approved):
👍

nickdenardis:
Should we at least have a reference to the `h1` tag should not be used since the page title is already the H1 on the page?

krystosch:
That would be helpful, just so they know it exists and that the style of it is already in place. @breakdancingcat - thoughts?

breakdancingcat:
@krystosch Yes, can you change it to something like h1 is the page title which is already on each page so headings for your page will be h2-h4.

PR #542, opened 2022-01-19
Update styleguide.blade.php

krystosch:
Updated H1 content. This will put an error on our page with out of order heading levels, but I'm OK with it since it's to show style. Thoughts? @nickdenardis


nickdenardis:
@krystosch I wonder if there is a way to note this without putting an H1 on the page. That way our pages stay compliant and we don't have anyone scanning quickly and think the H1 style is in place.

Possible to use the copy but without the h1 above it?

PR #545, opened 2022-01-26
Hotfix/7.0.7

chrispelzer:
- PeopleRepository: If the field is a file, then set a field as the url to match the current structure of Profiles @chrispelzer


nickdenardis (approved):
👍

PR #546, opened 2022-01-28
Hotfix/7.0.8

chrispelzer:
scope $accordion_page to $base['accordion_page'] since it is coming from the DataMiddleware in the `childpage` view


nickdenardis (commented):
👍

PR #547, opened 2022-02-07
Update to Tailwind v3 from v2

chrispelzer:
Since we were so close to release of Base v7 with Tailwind V2 and due to some issues with Tailwind V3 we held off until they were fixed.

Now that they had some hotfixes under V3, we tested it and it worked fine purging classes again.


nickdenardis (commented):
👍

PR #549, opened 2022-02-08
Hotfix/7.1.1

chrispelzer:
Removes `mode: 'jit'` not needed in the `tailwind.config.js` due to it being enabled by default, was left over from Tailwind V2


nickdenardis (commented):
👍

PR #550, opened 2022-02-11
use artisan test to display out the failed test quickly

chrispelzer:
Why not?

Before:
Screen Shot 2022-02-11 at 5 06 33 PM

After:
Screen Shot 2022-02-11 at 5 03 04 PM


nickdenardis (approved):
Looks much better to me! 👍

PR #558, opened 2022-03-07
Hotfix/7.1.8

chrispelzer:
When using the `base:feature`, two files require phplinting after creation, this corrects it so that it doesn't need to


breakdancingcat:
👍

nickdenardis (commented):
👍

PR #559, opened 2022-03-07
Hotfix/7.1.9

chrispelzer:
Match PageFactory used as other otherwise it causes an error on initial creation


nickdenardis (commented):
👍  I ran into this on a site the other day, thanks for investigating!

PR #560, opened 2022-03-16
Hotfix/7.1.10

chrispelzer:
Rework the way `base:feature` generates the Feature with the stabs to be more consistent in the singular form of the Feature name, rather than trying to handle pluralization.

If multiple of the Feature are returned which is the case of the inital creation we use `getFeatureItems` and `featureitems` to desginate the multiple


nickdenardis (commented):
👍  Great change!

breakdancingcat:
🥳 thank you!

PR #561, opened 2022-03-16
Hotfix/7.1.11

chrispelzer:
Issue when adding a new `base:feature` it wouldn't add the proper menu page id to the styleguide Page, which then resulted in the `all_styleguide_tests` to not cover any newly created features due to it always testing the site_specific page at 999


nickdenardis (commented):
👍

PR #571, opened 2022-06-30
Hotfix/7.1.15

chrispelzer:
append the md5 hash of the serialized api paramaters to not conflict with other api calls with the same prefix but different paramaters


nickdenardis (commented):
👍

PR #578, opened 2022-08-17
Feature/reverse paginate

chrispelzer:
Reverse the Paginate, due to a lot of the data already being returned by the News API in the meta to instead replace the current `setPaging` on the ArticleRepository


nickdenardis (commented):
👍

PR #580, opened 2022-10-13
Added surtitle to the title tag of the page.

tracewayne:
Added surtitle to the title tag of the page. Mainly used for School of Medicine sites to distinguish inside of search results.


nickdenardis:
This is looking good. Should this exclude sites that should not be displaying the main site?
https://github.com/waynestate/base-site/blob/master/config/base.php#L117

Typically when this option is disabled, it is because the main site name is the surname.

nickdenardis (commented):
👍

nickdenardis (approved):
Arg, approved now 🤦

PR #587, opened 2023-01-17
Logic for dropdown if only one profile group

Fw7424:
## Reason for change

When only one profile group is active, we want the dropdown to automatically be hidden. 
Multiple groups.
Screen Shot 2023-01-18 at 10 39 47 AM

Single profile group
Screen Shot 2023-01-18 at 10 48 32 AM


nickdenardis:
Can the screenshots be included with this ticket since this is public and the link you provided are not.

Also, can you add an explanation for this does directly in the description of the PR so it doesn't link off to a private link for context?

breakdancingcat (changes requested):
Remove extra lines added to profile controller

nickdenardis:
Since we are putting a number of things into the next release, let's move this to the `develop` branch instead of this being a hotfix.

PR #595, opened 2023-02-14
Fix Menu Repository 500 error

chrispelzer:
If no menu is returned by the API then set the menu array as blank rather than erroring as a 500, this will log the error to the error level which will get picked up for easy review


nickdenardis (commented):
👍

PR #596, opened 2023-02-14
Hotfix/7.1.20

chrispelzer:
Set up the return the single_group array to for the ProfileRepository and PeopleRepository and corrected tests to allow 100% coverage


nickdenardis (commented):
👍

PR #602, opened 2023-02-20
Clean up the ProfileRepository from the styleguide

chrispelzer:
get rid of styleguide ProfileRepository since it will always be using the styleguide PeopleRepository due to the AppServiceProvider binding to the PeopleRepository when on the styleguide, they both adhere to the ProfileRepositoryContract which doesn't matter with the current set up of the styleguide/repositories

phplinted


nickdenardis (commented):
👍

PR #604, opened 2023-02-27
Feature/forms dropdown arrows

AmbyrElan:
## Reason for change

- Add dropdown menu arrows 

## Reminders

- Duque Axe accessibility check
- Verified each page looks good on each viewport size
- Add an example to the styleguide if applicable
- Add tests to prove the code works
- Provide CMS site link if applicable

## Screenshots / Videos

| Before | After |
|--------|--------|
|old-base-wayne-edu-styleguide-forms-2023-02-27-16_17_23|new-updated-base-wayne-local-styleguide-forms-2023-02-28-12_55_21|


chrispelzer:
Are we getting a double 4 using both `right-4 mr-4`?

nickdenardis (changes requested):
The placement of the arrow is off. It is far too much to the left.

We're looking for an implementation that doesn't require an additional SVG icon or anything added to the input field.

We will have forms embedded into the site that we do not control and will not be able to manually add an SVG icon to every dropdown.

Bringing the dropdown arrow back should be done in CSS only since we can't edit the HTML of embedded forms.

breakdancingcat:
@AmbyrElan try to find the code in the formy scss that's disabling the browser arrow
https://github.com/waynestate/base-site/blob/master/resources/scss/components/_formy.scss#L59

breakdancingcat (changes requested):
@AmbyrElan Please change the code in the formy scss that is removing the browser arrow. We should use the browser default unless we have a better solution. (Also remove the code for IE)
https://github.com/waynestate/base-site/blob/master/resources/scss/components/_formy.scss#L59

breakdancingcat (approved):
Alright! Let's try it on dev. Thanks!

nickdenardis:
Yes, let us know when it is up and we can do an accessibility and functionality review.

PR #606, opened 2023-02-28
add faculty output for an article to the individual article view

chrispelzer:
add faculty output for an article to the individual article view this matches the same style as the assets for the article.

Both might need to be looked at more for styling.

Screenshot 2023-02-28 at 3 46 43 PM


nickdenardis (commented):
👍

breakdancingcat:
👍

PR #607, opened 2023-02-28
Feature/travis to GitHub actions

chrispelzer:
It looks like the last like the Coveralls was ran was back in 2018. Also Travis-CI as well.

This replaces what it used to do but using the Github Actions which are built into the repository.


nickdenardis:
👍

PR #609, opened 2023-03-02
Recent Articles (default 10) displays on the profile

chrispelzer:
If available the articles will be displayed with the profile, similar to how the Courses taught is

Screenshot 2023-03-02 at 1 59 07 PM


nickdenardis (commented):
👍

PR #613, opened 2023-03-07
feature/cms-layout

breakdancingcat:
## Reason for change

If there isn't enough text, the div doesn't expand to full width

## Screenshots / Videos

| Before | After |
|--------|--------|
|Screenshot 2023-03-06 at 8 39 04 PM|Screenshot 2023-03-06 at 8 38 56 PM|

### DIFF
--- a/styleguide/Views/styleguide-basic-layouts.blade.php
+++ b/styleguide/Views/styleguide-basic-layouts.blade.php
@@ -66,7 +66,7 @@
 
         <div class="two-col-layout">
             <div class="md:w-1/3 flex-shrink-0">
-                <img src="/styleguide/image/300x200" alt="Image left placeholder image">
+                <img src="/styleguide/image/600x400" alt="Image left placeholder image">
             </div>
             <div class="w-full">
                 <p>{{ $faker->text(250) }}</p>
@@ -76,7 +76,7 @@
 <pre class="code-block" tabindex="0">
 {!! htmlspecialchars('<div class="two-col-layout">
     <div class="md:w-1/3 flex-shrink-0">
-        <img src="https://base.wayne.edu/styleguide/image/300x200" alt="Your image description">
+        <img src="https://via.placeholder.com/600x400" alt="Your image description">
     </div>
     <div>
         <p>'.$faker->text(60).'</p>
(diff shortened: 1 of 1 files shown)


styleguide/Views/styleguide-basic-layouts.blade.php
```diff
@@ -66,17 +66,17 @@
...
+            <div class="w-full">
                 <p>{{ $faker->text(250) }}</p>
                 <p>{{ $faker->text(250) }}</p>
             </div>
         </div>
 <pre class="code-block" tabindex="0">
 {!! htmlspecialchars('<div class="two-col-layout">
     <div class="md:w-1/3 flex-shrink-0">
-        <img src="https://base.wayne.edu/styleguide/image/300x200" alt="Your image description">
+        <img src="https://via.placeholder.com/600x400" alt="Your image description">
```
nickdenardis, line 79: Is there a reason this isn't using the build-in offline placeholer image generator?

nickdenardis (commented):
Looks good, one inline comment.

styleguide/Views/styleguide-basic-layouts.blade.php
breakdancingcat: The style guide image is broken in the CMS. If you visit the style guide url out of the CMS, it's a bunch of characters. The CMS doesn't know what to do with it. https://base.wayne.edu/styleguide/image/600x400

chrispelzer: These are the 2 responses for the images in the CKEditor 4. It seems even though when going through the browser to view https://base.wayne.edu/styleguide/image/600x400 it still returns it as html instead of a png.

Which I can assume might need updating to work with the latest Laravel or `response(200)` regardless of using the header, it doesn't overwride it. https://github.com/waynestate/image-faker/blob/master/src/ImageFakerController.php#L42

Styleguide image:
Screenshot 2023-03-07 at 6 59 23 AM

Placehold-it image
Screenshot 2023-03-07 at 6 59 18 AM

chrispelzer: @nickdenardis @breakdancingcat 

Nevermind, it's due to the CMS not being the same URL as base.wayne.edu or whatever site we are getting that styleguide image off of.

This 404s it, if it's not on the site's URL. https://github.com/waynestate/image-faker/blob/master/src/ImageFakerController.php#L29-L31

nickdenardis: This has been resolved:
https://github.com/waynestate/image-faker/pull/4

It will be in Base after the next dependency update.

PR #620, opened 2023-04-19
Feature/back to profile base config array

Fw7424:
Create an array of site ids to define the back url


nickdenardis:
I'm wondering what this solves for 80% of sites. This feels like a solution for a few rare occasions.

nickdenardis (changes requested):
Two outstanding questions:

1. Does this solve the "go back to profiles" url issue in all situations?
2. What does this solve for 80% of sites? This feels like a solution for a few rare occasions.

PR #622, opened 2023-04-19
Feature/external link icon class

Fw7424:
* External link class can be added to the menu item "class" field to display an accessible external link icon at the end of the menu item name. 
* WIP external link class for under menu items. The class itself works but we do not have a way to enable it from the CMS. 
Screenshot 2023-07-28 at 11 48 59 AM
Screenshot 2023-07-28 at 11 43 26 AM
Screenshot 2023-07-28 at 11 50 31 AM


nickdenardis:
Can we get some more insights into the visual of this? 
How does this look in various situations and on small screens?

nickdenardis (approved):
Looks good to me!

PR #637, opened 2023-07-14
Replace the site title H1 to ensure there is only one primary H1 on a page

nickdenardis:
## Reason for change

SEO enhancement. This upgrades the [wsuheader package](https://github.com/waynestate/wsuheader/pull/54) and removes the h1 tags for the site title. This results in only a single H1 tag on the site for the page title.

## Demo

| Before | After |
|--------|--------|
| Screenshot 2023-07-14 at 3 24 05 PM | Screenshot 2023-07-14 at 3 41 24 PM |

| Before | After |
|--------|--------|
| Screenshot 2023-07-14 at 3 24 27 PM | Screenshot 2023-07-14 at 3 41 15 PM |

| Before | After |
|--------|--------|
| Screenshot 2023-07-14 at 3 41 50 PM | Screenshot 2023-07-14 at 3 41 35 PM |

| Before | After |
|--------|--------|
| Screenshot 2023-07-14 at 3 50 14 PM | Screenshot 2023-07-14 at 3 50 27 PM |

### DIFF
--- a/resources/views/components/header.blade.php
+++ b/resources/views/components/header.blade.php
@@ -1,7 +1,7 @@
 <div class="wsuheader">
     <div class="wsuheaderwrap">
         <div class="wsuwordmark">
-            <h1>
+            <div>
                 <a href="https://wayne.edu/" aria-labelledby="wsuheader-title">
                     <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 605 65" preserveAspectRatio="xMaxYMin meet">
                         <title id="wsuheader-title">Wayne State University</title>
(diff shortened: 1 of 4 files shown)


breakdancingcat:
The titles in the heroes aren't bolded anymore after this change. This might be okay though, I'll determine if it should be switched back after it gets deployed.

nickdenardis:
> The titles in the heroes aren't bolded anymore after this change. This might be okay though, I'll determine if it should be switched back after it gets deployed.

I'm good with this not being bold as the text is already so large.

PR #640, opened 2023-08-02
Feature/external icon pt 2

breakdancingcat:
Links within the content, main-menu and under-menu classes that contain https://, http://, and // will automatically have an external link icon applied. This applies to text links as well as buttons. Buttons and links outside of the conditions above will not be modified.
### Styleguide
base-external-icon-styleguide
### Content area
base-external-icon-content-area
### Under menu
base-external-link-under-menu


factories/UnderMenu.php
breakdancingcat, line 1: So the external link icon isn't showing by default on the under menu styleguide page

nickdenardis:
This may be a little weird since although these are going outside of the current site there are plenty of reasons to do this on our sites due to the subdomain setup. 

Traditionally, this icon in content is limited to links opening in a new window. Can this be limited to body content has the `target="_blank"` property? 

Otherwise, we would have to do more use case testing with real content before putting this in for any 'http...' links in page content.

breakdancingcat:
@nickdenardis I believe we are recommending people not to use target="_blank". The goal from our content team was to provide context where users are being redirected off of the current s/c/d page. This is scoped only to .content, .main-menu, and .under-menu. I tested it pretty thoroughly but I am interested in another site where we could add this to test.

nickdenardis (changes requested):
Let's make this external link designation available to the `.content` area but have it be opt-in only. Requiring the content owner to make the conscious decision to add the `external` class to an anchor tag that based on the context would be helpful for the end-user to know the link will be taking them to an external resource.

resources/scss/components/_external-icon.scss
```diff
@@ -0,0 +1,52 @@
...
+        @apply hidden;
+    }
+
+    &::after {
+        @extend .external-icon;
+    }
+}
+
+// Strictly inline anchors
+.content a.external-link {
```
nickdenardis, line 31: For consistency, let's continue to only use the class `.external`:
https://base.wayne.edu/styleguide/menu/external

breakdancingcat: Please explain how I need to be consistent with something that is a new idea? I suggest the specificity of external-link, so if you mean changing the styleguide url to match I can do that.

breakdancingcat:
@nickdenardis I've made the requested changes, can you let me know if you are good with these changes?

nickdenardis:
> @nickdenardis I've made the requested changes, can you let me know if you are good with these changes?

Looks good! Only one comment added inline.

nickdenardis (approved):

PR #643, opened 2023-08-17
Feature/button adjustments

breakdancingcat:
When working through how the external icon appears on buttons, I found old classes and old formatting that needed freshening up

|Before|After|
|--|--|
|||


nickdenardis (approved):
🔥

PR #651, opened 2023-09-18
Hotfix/8.4.6

chrispelzer:
Small fix to the PromoPage instead of making raw json, we can use json_encode and encode the PHP array


nickdenardis (commented):
👍

PR #653, opened 2023-09-28
Hotfix/8.4.8

chrispelzer:
Fixes issue with news articles not redirecting when linking directly to the article.


nickdenardis (commented):
👍

PR #656, opened 2023-10-17
Feature/modularpage

breakdancingcat:
- Testing out modular page with the CMS.
- Repository collects all page fields prefixed with "modular" and uses the label to append a component filename to each promo item to access in the front end. 
- Page field has to match component name, need to check that 'file exists' or handle invalid component name in some way.
- Components are in the modular folder.


.phpunit.cache/test-results
```diff
@@ -0,0 +1 @@
...in_contact_and_under_menu":0,"Tests\\Unit\\Repositories\\PromoRepositoryTest::subsite_contact_merging_with_no_main_contact":0.001,"Tests\\Unit\\Repositories\\PromoRepositoryTest::subsite_under_menu_merging_with_no_main_under_menu":0.001,"Tests\\Unit\\Repositories\\PromoRepositoryTest::getting_homepage_promos_should_return_array":0,"Tests\\Unit\\Repositories\\TopicRepositoryTest::getting_topics_should_return_array_of_topics":0.001,"Tests\\Unit\\Repositories\\TopicRepositoryTest::getting_topics_with_exception_should_return_empty_array":0,"Tests\\Unit\\Repositories\\TopicRepositoryTest::finding_topic_should_return_topic":0,"Tests\\Unit\\Repositories\\TopicRepositoryTest::sorting_topics_by_letter_should_be_in_order":0,"Tests\\Unit\\Repositories\\TopicRepositoryTest::setting_selected_topic":0,"Tests\\Unit\\Repositories\\TopicRepositoryTest::setting_no_selected_topic":0,"Tests\\Unit\\Support\\HelpersTest::merge_string_value_should_throw_exception":0,"Tests\\Unit\\Support\\HelpersTest::merge_array_with_zero_key_should_throw_exception":0,"Tests\\Unit\\Support\\HelpersTest::merge_array_with_same_key_should_throw_exception":0,"Tests\\Unit\\Support\\HelpersTest::merge_should_merge_arrays":0}}
```
nickdenardis, line 1: Let's add this directory to the `.gitignore` file.

app/Http/Controllers/ModularPageController.php
```diff
@@ -0,0 +1,47 @@
...
+     * Display the homepage view.
+     *
+     * @param Request $request
+     * @return \Illuminate\View\View
+     */
+    public function index(Request $request)
+    {
+        $promos['promos'] = $this->promo->getModularPromos($request->data['base']);
+
+        //$articles = $this->article->listing($request->data['base']['data']['news_application_id'] ?? $request->data['base']['site']['news']['application_id']);
```
nickdenardis, line 40: Perhaps we should mark all temporary comments/code with `TODO: ` so we can keep track of and find these later to ensure none make it into a release.

app/Repositories/ModularPageRepository.php
```diff
@@ -0,0 +1,149 @@
...
+
+        // Append actual page id to config
+        if (str_contains($component['config'], 'page_id')) {
+            $component['config'] = preg_replace('/\bpage_id\b/', 'page_id:'.$page_id, $component['config']);
+        }
+
+        // Not allowing "first" config option
+        // API seems to do fine with double pipes, so not handling them at this time
+        if (str_contains($component['config'], 'first')) {
+            $component['config'] = preg_replace('/\bfirst\b/', '', $component['config']);
```
nickdenardis, line 122: This feels safe to do without the conditional around it. Is there a config situation where config options would have the chars `first` in it that are not prefixed or suffixed by a pipe? 
https://github.com/waynestate/parse-promos/blob/master/README.md#usage

factories/GenericPromo.php
```diff
@@ -24,7 +24,7 @@ public function create($limit = 1, $flatten = false, $options = [])
     {
         $promo_group_id = $this->faker->randomNumber(5);
 
-        for ($i = 1; $i <= $limit; $i++) {
+        for ($i = 0; $i <= ($limit - 1); $i++) {
```
nickdenardis, line 27: Small thing here, this could be done without the internal parenthesis:
`for ($i = 0; $i < $limit; $i++) {`

resources/views/modular/components/button-column.blade.php
```diff
@@ -0,0 +1,8 @@
+<div>
```
nickdenardis, line 1: Small thing, but let's remember to have the params commented at the top of these components to match the others.

resources/views/modular/components/spotlight.blade.php
```diff
@@ -0,0 +1,24 @@
...
+    <blockquote class="relative flex border-gold border-l-12 pt-4 lg:pl-8 mx-auto my-0">
+        <div class="flex flex-wrap items-center w-full">
+            <div class="content w-full text-lg md:text-xl lg:text-xl lg:leading-relaxed">{!! $data['description'] !!}</div>
+            <div class="w-full flex items-start gap-x-2 mb-4 lg:mb-2">
+                <div class="gold-half-circle-bg w-24 lg:w-1/4 lg:absolute top-0 right-0">
+                    @image($data['relative_url'], $data['filename_alt_text'], 'block rounded-full relative z-10')
+                </div>
+                <div>
+                    <h2 class="mb-0 text-lg xl:text-xl">{{ $data['title'] }}</h2>
+                    <cite class="not-italic xl:text-lg">{{ $data['excerpt'] }}</cite>
```
nickdenardis, line 16: Is the `title` or `excerpt` the name of the person who is providing the quote?

Want to make sure `<cite>` surrounds their name, and potential context (Transfer student, program name, etc).
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/cite

resources/views/modular/components/steps-column.blade.php
```diff
@@ -0,0 +1,20 @@
...
+    <ul class="grid grid-cols-1 gap-6">
+    @foreach($data as $step)
+        <li>
+            <a class="block group" href="{{ $step['link'] }}">
+                <div class="flex gap-3">
+                    <div class="w-16 grow-0 shrink-0">
+                        @image($step['relative_url'], $step['filename_alt_text'], 'block mx-auto')
+                    </div>
+                    <div>
+                        <h3 class="text-lg lg:text-base xl:text-xl mt-0 mb-1 underline group-hover:no-underline">{{ $step['title'] }}</h3>
```
nickdenardis, line 12: Should this also have the `h2`/`h3` conditional if there is no component heading so it doesn't skip to `h3`?

nickdenardis (commented):
Looking good! Only a few inline comments/questions.

app/Repositories/ModularPageRepository.php
breakdancingcat: @nickdenardis The issue is which pipe to remove, we can't be sure someone won't put it at the beginning or end of the config. Example `randomize|page_id|first` is valid, or where the only config option is `first` without any pipes. If anyone were to accidentally use first, it would break the page, so I have to be sure its removed.

PR #660, opened 2023-10-23
Feature/modular 2

breakdancingcat:
* Moved the events/articles to a modular component and removed them from the controller
* Carried over "change promo item display" from promo page (removes excerpt, description from showing, appends a single promo view page link) 
* Fixed the spotlight component, and  fixed stlying for video row, video column, image promo
* Chris is adding a calendar url to the event api so that's what $dates[0]['calendar_url'] is for (Reducing the need for a separate page field for calendar url) 
* Rearranged styleguide page


nickdenardis (commented):
👍

PR #661, opened 2023-10-24
Feature/social image override

chrispelzer:
if social_image is available then use that

add coverage and tests to make sure social_image is available


nickdenardis (approved):
👍

PR #666, opened 2023-10-26
tweaks for the modular styleguide page using the styleguide modular repository

chrispelzer:
remove unneeded promoPage data

update modular-news and modular-event page field names

update styleguide ModularPage repository to adhere to the component information not being underneath each item


nickdenardis (commented):
👍

PR #676, opened 2023-11-20
Feature/button updates

breakdancingcat:
## Refactoring how the button component works
* Replaced 'under-menu' component with 'button-column', no duplication of code and requires the same functionality
* Reduced image button options that were not used frequently or that I no longer want to support (Background image with a white or black overlay and text on top)
* Combined under menu and button column information
* Added PDF overlay for image buttons
* Instead of the promo item option passing in the filename, it passes in a class for the same shared button file, so i removed several files of essentially the same code. If that class name doesn't exist, the button appears default gray. You can pass in all of the button options like default, green, green gradient, gold, gold gradient, image. The CMS will be updated to only have "Default, green, image" as the default options, the others will have to be added by one of us if the client requests more options.

image


nickdenardis (approved):
👍

PR #677, opened 2023-11-27
Limit page title to 60-70 chars for SEO

nickdenardis:
## Reason for change

When page titles are longer, the addition of the site name, sur-title name, and then "Wayne State University" pushes the title tag into the triple digits for length.

SEO recommends 60-character limit for page titles. 

## Proposed solution

~~Because the page title can be changed up to the line before the view is called, the logic needs to exist in the template itself. Dropping right into raw PHP in Blade.~~

Moved the logic to a helper function which is called within the `merge()` function before the view.

Tests have also been added.

- The page title from the CMS is not modified and is prioritized
- The sur-title site name has been dropped, it isn't needed in this context since it is already visible on the page
- The site title is optional based on space. From all the recommendations we have seen, this is less important than the university name
- University name is included only if there is enough room and doesn't push it over an extended 70-char limit

## Examples

https://engineering.wayne.edu/mechanical/academics/bs
Current: `<title>Bachelor of Science in mechanical engineering - Mechanical Engineering - Wayne State University</title>`
Ideal: `<title>Bachelor of Science in mechanical engineering - Wayne State University</title>`

https://engineering.wayne.edu/electrical-computer/academics/master-electrical-engineering
Current: `<title>Master of science in electrical engineering - Electrical and Computer Engineering - Wayne State University</title>`
Ideal: `<title>Master of science in electrical engineering - Wayne State University</title>`

## TODO

- [x] Logic in place
- [x] Tests in place


breakdancingcat (approved):
👍🏻 Thank you!

chrispelzer (commented):
Works for me, lot cleaner

PR #680, opened 2023-12-07
Feature/modular updates

breakdancingcat:
* Added back logic to fix erroneous JSON by removing all spaces/line breaks, and removing a trailing comma.
* Moving modular components to the childpage instead of it's own template for wider adaptability (Deleted ModularPageController).
* Added legacy support for `accordion_promo_group_id`, `listing_promo_group_id`, and `grid_promo_group_id` so these page fields turn into their respective components with default settings on any childpage. 
* Renamed “Modular docs” to “Using components” page and moved it into the components menu.
* Updating the styleguide component pages to be a better structured and more informative resource.
* Added functionality to organize the catalog component by option and deleted the "PromoPageController".
* Moved the individual promo view from promo page to "PromoController".
* Moved "back to promo page" function to the PromoRepository from the PromoPageRepository.
* Added a component to display a single promo item column and single promo item row (because Spotlight is for a quote with a citation).
* Moved the "More news" link into the conditional if news data exists.
* Added a spotlight column component (previously only a row).
* Created a HomepageRepository and moved the GetHomepagePromos function from the PromoRepository to it.
* Styleguide updates and fixes.
* Test updates and fixes (100% coverage).


nickdenardis (commented):
👍

PR #684, opened 2023-12-19
Moving modular template configuration to childpage and making components consistent

breakdancingcat:
* Moved modular template logic to childpage and remove modular template page/controller
* Updated components to reflect their specific options and have a consistent view across the board 
* Added a new childpage controller test to cover modular template logic


nickdenardis (approved):
👍

PR #685, opened 2024-01-10
Release/8.7.0 - Moving modular template configuration to childpage and making components consistent

breakdancingcat:
* Moving modular template configuration to childpage and making components consistent
* Optimizing components to mirror each other as much as possible and reduce duplicate information


nickdenardis (approved):
👍

PR #694, opened 2024-02-05
Hotfix/8.8.1

chrispelzer:
Subsite folder is necessary for the subsite to be appended to the topic url, otherwise it will cause confusion when navigating between listing and individual view when on a subsite


nickdenardis (commented):
👍

breakdancingcat:
Thank you!

PR #696, opened 2024-02-16
hotfix/8.8.4 - Need base version in the header on sites, not 'version'

breakdancingcat:
After upgrading another site and seeing that site's version in the header and not the version of Base, I realized I needed 'base version' not 'version'


nickdenardis (approved):
👍

PR #698, opened 2024-02-20
Feature/extending-repository

breakdancingcat:
Created the framework for extending the promo repository for global promos, so we do not have to edit the promo repository file and upgrades are cleaner.


chrispelzer (approved):
Yea, this will get the point across

PR #707, opened 2024-04-18
refactor: switch the git hook pre-commit to use docker instead of a wild wayne

chrispelzer:
Due to switching from a Vagrant environment to Docker we can switch the pre-commit now


nickdenardis (commented):
👍

PR #710, opened 2024-04-30
Hotfix/8.10.1

chrispelzer:
Allows both .localhost and .local to be used.

[fix: add localhost as a local option within the .htaccess](https://github.com/waynestate/base-site/commit/40ccb42d89aa43178cab8167c0ad34fb1237189e)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "base",
   "private": true,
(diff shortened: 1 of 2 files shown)


nickdenardis (commented):
👍

PR #711, opened 2024-05-01
Feature/component party

breakdancingcat:
* Moved component loop into its own partial
* Created a specific styleguide accordion to more easily display component details (only used on the accordion controller at the moment) 
* Organized styleguide by renaming files to follow a new naming convention
* Removed some descriptive component text that is better suited for Canvas


nickdenardis (approved):
👍

PR #712, opened 2024-05-02
Feature/accordion upgrade

breakdancingcat:
- Adding the ability to upload an image with a caption(excerpt) to an accordion item.
- Options are available to change the orientation from default right, to left, and center(below)
- Debating adding in italic Lato into Base by default, when it is used it looks much nicer than browser italic, and we use it in the figure component so this wouldn't just be for accordion
- One downside, these images cannot be lazy loaded because of the accordion js detecting the height 
- Adapted from the Business site 

Screenshot 2024-05-01 at 11 36 50 PM
Screenshot 2024-05-01 at 11 35 56 PM


nickdenardis (approved):
👍

PR #713, opened 2024-05-02
Feature/flag upgrade

breakdancingcat:
When moving the Flag component back into the component menu, I remembered some improvements that needed to be made 
- The flag can now expand to fit more text within it
- Used a polygon instead of borders to create the bottom triangle which allows it to expand if needed
- Included the ability to use em in the title tag instead of having to separate "Make a" and "gift" between title/excerpt. Excerpt is still supported but not necessary. This also allows users to make the entire title em Larger italic if needed 
- (This did not take long and it was a pet peeve)

Screenshot 2024-05-01 at 11 51 53 PM

Obviously we wouldn't approve this, but I wanted to demonstrate it's responsiveness where before it would break to a new line and be all jacked up looking
Screenshot 2024-05-01 at 11 47 00 PM

Screenshot 2024-05-01 at 11 47 22 PM

Screenshot 2024-05-01 at 2 11 39 PM

Before 
Screenshot 2024-05-01 at 11 50 20 PM


nickdenardis (approved):
👍

PR #714, opened 2024-05-02
Release/8.10.2

breakdancingcat:
- Moved the component loop to it's own partial 
- Made components global
- Set up an easier way to outline component configuration details with a new styleguide accordion component 
- Organized styleguide files with naming convention prefixes


nickdenardis (commented):
👍

PR #716, opened 2024-05-28
Hotfix/8.10.3

chrispelzer:
fix: fix error `bash: cannot set terminal process group (-1): Inappropriate ioctl for device` when running docker exec command, faketty() is required due to sourcetree erroring with -it to fake the terminal

```
❯ docker exec -i wsu-base bash -ilc "make phplintdry"
bash: cannot set terminal process group (-1): Inappropriate ioctl for device
bash: no job control in this shell
```

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,4 +1,37 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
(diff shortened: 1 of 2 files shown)


nickdenardis (commented):
👍

PR #723, opened 2024-08-06
Canonical URL for individual news articles based on permalink

nickdenardis:
## Reason for change

Because articles can be renamed and potentially accessed at different URLs, the Canonical URL assumes the current URL is the most up-to-date and uses that URL as the canonical. 

This causes a situation where the same article could be accessed at multiple URLs, each with its own canonical URL. Search engines complained the same content was pointing to different canonical URLs. 

This change forces the most up-to-date title (slug) to be used as the canonical URL so search engines index the correct URL.

## Reminders

- Styleguide example in place

## Demo

| Before | After |
|--------|--------|
| Screenshot 2024-08-06 at 7 20 39 AM | Screenshot 2024-08-06 at 7 21 07 AM |


breakdancingcat:
👍

PR #740, opened 2024-10-08
fix: Setting the base page controller data to array.

Fw7424:
issue with hero page info coming in as string when array is expected

This issue shows up when trying to contained hero pages to full-width via modular component.


breakdancingcat (approved):
Thank you!

nickdenardis (commented):
This looks good. 

Not dealbreakers at this point, but ideally:

1. Possible to add the before and after screenshots?
2. Is there a test we can add to prevent regressions in the future?

PR #741, opened 2024-10-08
Upgrade stylelint from v13 to v16

chrispelzer:
Fixes #634

Stylelint now includes a `--fix` option similar to phplinting so it is able to correct the errors without much issue

`stylelint ./resources/scss/**/*.scss --fix`

PR #735, opened 2024-09-30
Feature/contain breadcrumbs

Fw7424:
Setting the breadcrumbs to be contained when banner flag is present. this issue occurred on CLAS https://bitbucket.org/waynestate/clas/pull-requests/340/overview.

Below the example from clas with a before and after.
3497721042-Screenshot 2024-06-28 at 9 58 14 AM
2060587443-Screenshot 2024-06-28 at 10 28 22 AM

PR #734, opened 2024-09-26
base:module command line now in place

nickdenardis:
## Reason for change

As more functionality on sites is moving to the modular components instead of full templates, this change introduces the `php artisan base:module {name}` command line interface to generate the scaffolding for new modular components.

It works similarly to the `base:feature` and introduces a new "Site specific" menu item in the styleguide to ensure the site-specific modular components are isolated and do not slow down future codebase upgrades.

The stub component intentionally does not do anything other than dump the promotion items and config to the screen. 

The "Site specific" styleguide page walks through the process of creation and the next steps to update the styleguide pages with description, config and instructions.

## Reminders

- Update package version number
- Frontend accessibility check
- Styleguide example in place
- New functionality covered by tests
- Screenshots of multiple screen sizes

## Demo

Screenshot 2024-09-26 at 6 35 13 AM

| Styleguide before | Styleguide after |
|--------|--------|
| Screenshot 2024-09-26 at 6 35 48 AM | Screenshot 2024-09-26 at 6 36 32 AM |

PR #732, opened 2024-09-24
feature/headings - Dividers - adjust padding on heading dividers

AmbyrElan:
# Updates 

## Reason for change 
- Moving down the heading bars so they appear more centered with the heading text 

|Before|After|
|-----|-----|
|image|image

PR #742, opened 2024-10-10
Release 8.12.0 - Profile titles, base:module command, heading dividers and breadcrumb spacing

nickdenardis:
### [Allow line breaks only in the Title profile field](https://github.com/waynestate/base-site/pull/738)

We are already allowing linebreaks in the Title field on the main university directory, we should allow it on the individual sites to match.

This change allows `<br>` tags only in the Title field.

### [base:module command line now in place](https://github.com/waynestate/base-site/pull/734)

As more functionality on sites is moving to the modular components instead of full templates, this change introduces the `php artisan base:module {name}` command line interface to generate the scaffolding for new modular components.

It works similarly to the `base:feature` and introduces a new "Site specific" menu item in the styleguide to ensure the site-specific modular components are isolated and do not slow down future codebase upgrades.

The stub component intentionally does not do anything other than dump the promotion items and config to the screen.

The "Site specific" styleguide page walks through the process of creation and the next steps to update the styleguide pages with description, config and instructions.

### [Dividers - adjust padding on heading dividers](https://github.com/waynestate/base-site/pull/732)

Updates the position of the header dividers to be more aligned to the center of the words. 

### [Prevent breadcrumbs from running into the Flag](https://github.com/waynestate/base-site/pull/735)

Updates to the placement of the breadcrumbs and flag component. They no longer run into each other.

### [Upgrade stylelint from v13 to v16](https://github.com/waynestate/base-site/pull/741)

Maintenance update for this dependency.

### [Setting the base page controller data to array.](https://github.com/waynestate/base-site/pull/740)

Ensuring the `hero` modular component correctly updates the full width controller 

### [Default back to profiles link now returns 200 response](https://github.com/waynestate/base-site/pull/737)

Updates the URL structure of the default "profile_return_url" from "/profiles/" to "/profiles" because we automatically remove the trailing slash. This reduces an unnecessary redirect.

### [Individual profile name now showing correctly as the page title](https://github.com/waynestate/base-site/pull/736)

Fixes a bug that prevented the full profile user name on the individual profile view. This was introduced in 8.11.0 and is now corrected.

### [Tests passing again after generating a new feature](https://github.com/waynestate/base-site/pull/733)

Due to the changing structure of the childpage, after generating a `base:feature` the tests would immediately fail. This change resolves that issue.


chrispelzer:
I added the `make stylelintfix` even though it seems odd. Our current set up is using hooks with the dry being `make stylelint` which our hook is using `make phplintdry`

This keeps it in line with our phplint and hooks.
https://github.com/waynestate/base-site/blob/master/hooks/pre-commit#L68

nickdenardis:
@chrispelzer 
Let's keep it to these two commands so it matches the others:

```
make stylelintdry 
make stylelint
```

`make stylelintdry` - Doesn't have the --fix
`make stylelint` - Does have --fix 

This does then change the default behavior of `make stylelint` but it makes it consistent with the other commands.

nickdenardis:
@chrispelzer 
If it affects the user experience of the pre-commit hooks and makes that more difficult with auto-fix, then we can move it to the `make stylelintdry` to warn someone of needs before the commit. But we'd have to see how much of an issue/confusion fixing it precommit would cause.

chrispelzer:
@nickdenardis I responded in basecamp since it is a little more indepth and across other areas

breakdancingcat:
👍🏻

PR #745, opened 2024-11-04
fix: breadcrumbs menu flag variable

AmbyrElan:
## Description

- Noticed that the variable banner needs to be updated to flag in the [breadcrumbs component](https://github.com/waynestate/base-site/blob/master/resources/views/components/breadcrumbs.blade.php#L5) since it is being [referenced as flag](https://github.com/waynestate/base-site/blob/master/config/base.php#L277) now.


nickdenardis:
I wonder why/how this wasn't caught by the tests.

https://base.wayne.edu/styleguide/component/flag

Does this mean on current sites the flag is not displaying when it should be selected on a page?

breakdancingcat (approved):
Ty!

AmbyrElan:
> I wonder why/how this wasn't caught by the tests.
> 
> https://base.wayne.edu/styleguide/component/flag
> 
> Does this mean on current sites the flag is not displaying when it should be selected on a page?

@nickdenardis ,

- I checked out the 'develop' branch on Base v8.12. where the flag variable is still set to 'banner.' All the tests passed because the[ !empty check is around $base['banner'] in the breadcrumbs component](https://github.com/waynestate/base-site/blob/8.12.0/resources/views/components/breadcrumbs.blade.php#L5) and 'flag' is referenced everywhere else in the code including the styleguide.

- When I removed the !empty check around $base['banner'] on develop, the PageRepositoryTest threw two styleguide errors: 
[all styleguide routes should load successfully](https://github.com/waynestate/base-site/blob/8.12.0/tests/Feature/PageRepositoryTest.php#L15)
[all styleguide routes with no data should load successfully](https://github.com/waynestate/base-site/blob/8.12.0/tests/Feature/PageRepositoryTest.php#L26)
-  The change seem was applied in the [8.12.0 Base release](https://github.com/waynestate/base-site/blob/8.12.0/resources/views/components/breadcrumbs.blade.php#L5) and wasn't implemented yet on the [previous 8.11.2 version of Base](https://github.com/waynestate/base-site/blob/8.11.2/resources/views/components/breadcrumbs.blade.php#L5).  All the flags should work correctly on sites that have not yet been updated to 8.12.0.

cc, @breakdancingcat

breakdancingcat:
> I wonder why/how this wasn't caught by the tests.
> 
> https://base.wayne.edu/styleguide/component/flag
> 
> Does this mean on current sites the flag is not displaying when it should be selected on a page?

This only affects a class applied to the breadcrumbs when the flag is present, it wasn't caught by tests because it was only checking if the data exists and didnt use flag data for anything.

Commented in Basecamp too with screenshots, cc @nickdenardis @AmbyrElan

chrispelzer:
@breakdancingcat Currently right now we don't have real front end tests that would make sure that the class is applied.

It's all possible, which would be interesting cause we'd be able to have it run against /styleguide without an issue with rules that we want.

PR #746, opened 2024-11-06
Release 8.13.0 - Laravel 11 upgrade

nickdenardis:
## Dependency updates

[Updated codebase to Laravel 11](https://laravel.com/docs/11.x/upgrade).

We used [Laravel Shift](https://laravelshift.com/) to update the code automatically and it supplied us with changes it could not do on its own. This process saved at least 10 hours of developer time.

The result is a slimmer codebase (all default configuration files have been removed) the defaults are now built into the framework codebase and only changes need to be added to the application code.

## Transition

With this change, it means more items are defined directly in the .env file. To upgrade sites, we will be rolling out new .env updates before upgrading the site to ensure our unique configuration values are set on deployment.

## Features

There are no user-facing feature updates in this release.


AmbyrElan (approved):

AmbyrElan:
@nickdenardis ,

In reference to the dependency updates where '(all default configuration files have been removed) the defaults are now built into the framework codebase and only changes need to be added to the application code,' does this effect how we assign config items like controllers and global site promotions? Would we no longer use the config.base file to handle those assignments and handle those adjustments in the template controller instead?

nickdenardis:
> @nickdenardis ,
> 
> In reference to the dependency updates where '(all default configuration files have been removed) the defaults are now built into the framework codebase and only changes need to be added to the application code,' does this effect how we assign config items like controllers and global site promotions? Would we no longer use the config.base file to handle those assignments and handle those adjustments in the template controller instead?

@AmbyrElan It does not affect our code; it only affects framework-level configurations.

The base.php file has not been modified in this upgrade:
https://github.com/waynestate/base-site/blob/master/config/base.php

AmbyrElan:
> > @nickdenardis ,
> > 
> > In reference to the dependency updates where '(all default configuration files have been removed) the defaults are now built into the framework codebase and only changes need to be added to the application code,' does this effect how we assign config items like controllers and global site promotions? Would we no longer use the config.base file to handle those assignments and handle those adjustments in the template controller instead?
> 
> @AmbyrElan It does not affect our code; it only affects framework-level configurations.
> 
> The base.php file has not been modified in this upgrade:
> https://github.com/waynestate/base-site/blob/master/config/base.php

Thanks for clarifying!

PR #755, opened 2024-11-22
Hotfix/8.13.1 - CACHE_STORE configuration issue

chrispelzer:
- fix: CACHE_STORE should also be set to redis otherwise it conflicts with the CACHE_DRIVER not having the correct configuration
- various dependabot fixes


nickdenardis (commented):
👍

PR #758, opened 2024-12-02
Feature/profile config move to base config

Fw7424:
Instead of having to use random config options, we should use a single JSON config.


nickdenardis (approved):
Looks good to me 👍

breakdancingcat (approved):
Thanks Olu!

Fw7424:
not committing, I added changes for the styleguide information, I also refactored the parseProfileConfig to allow for greater flexibility.

PR #759, opened 2024-12-03
Feature/profile config move to base config

Fw7424:
This is a new PR for changes to profiles, 
This change will allow for profile data to be parsed from a json data in custom fields, similar to Modular components. 
changes to the variable names under profile, to keep the input for the user simple, and allow for additional variables to be added without need to adjust the code for parsing that input. 
Screenshot 2024-12-03 at 5 20 48 PM


nickdenardis (commented):
If we can get the one line not covered addressed, I'm good with this.

Fw7424:
@nickdenardis I will need some assistance in figuring out why this line is no longer covered. I've reached out to Chris who will be assisting me on Friday.

nickdenardis (approved):
Looks good to me 👍

PR #760, opened 2024-12-09
Remove PeopleRepository and replace with ProfileRepository as only repository

chrispelzer:
PeopleRepository is just causing more confusion due to us using it for the styleguide only area while also still having ProfileRepository/PeopleRepository on the non-styleguide.

This corrects it and just removes PeopleRepository completely and rely just on ProfileRepository but due to inconsistencies between the two we need to correct it overall for the styleguide.


breakdancingcat:
Thanks!

nickdenardis (commented):
Looks good to me! 👍

PR #767, opened 2024-12-19
Feature/modular events filter by title - Update the Base modular events component to display by title and accept a custom heading

AmbyrElan:
# Reason for change:

- Add unit tests for getEventsByTitle function:
    - getting events by title should return only events matching string array
        - Checks that each event returned has an event title matching the strings passed into the filterTitle key
    - getting events by title with no titles should return empty array
        - Checks that the method returns an empty array if no titles are passed into the filterTitle key
    - getting events by title should return unique events 
        - Checks that the method returns the expected events and verifies that the event ids are not duplicated to ensure repeating events are not showing up multiple times 
- Refactored unit tests for getEventsFullListing function:
    - Updated the function to key by date so that it works with the component updates.  getEvents and getEventsByTitle are both returned keyed by date, updating the  grtEventsFullListing function to key by date allowed me to minimize the code for the four events components so that I don't need a key to check if the events data is filtered or not
    - getting events full listing and getting events full listing no image were upated to accept the new data output
- Add new function, getEventsByTitle, to the EventsRepository class that allows filtering of the modular events component data based on one or more parsed title strings independent of case or whitespace
- Call new function in the ModularPageRepository class and update conditions to call the required EventsRepository class method based on if the filterTitle key has input
- Add check for custom title used on the events component and ensure that events is included in the heading
- Add the new filterTitle key to the component configuration example shown on the [Events listing component page](https://base.wayne.edu/styleguide/component/eventlisting) showing how to input a custom heading and filterTitle string 
- Add the filterTitle key definition to the [Using components JSON data explained](https://base.wayne.edu/styleguide/component/guide) page section

|Before|After|
|-------|-----|
|image|image|
|image|image|
|filterTitle is a new key so there is no before|image|


AmbyrElan:
@nickdenardis , @breakdancingcat ,

The getEvents, getEventsFullListing, and getEventsByTitle functions seem similar enough that it could become one function with $titles as an optional parameter utilizing the calendar.events.fulllisting api.  Is this a reasonable change to incorporate with this update?

There is a coverage decrease in the ModularEventsRepository function from coveralls. I will update this to a draft PR to investigate and update the code coverage.

nickdenardis:
@AmbyrElan 
After looking this over, it feels like the filter by title can be a second step instead of recreating the entire event fetch process.

Let's walk through this more before we commit to a specific implementation.

AmbyrElan:
* Nick and I reviewed the events repository functions along with the different events api calls used in each function and determined it would fit the goal of reducing the number of events repository class functions by utilizing a pre-existing search query present in the fulllisting api call. 
* The query doesn't currently have the functionality to accept an array of strings to search by like the getEventsByTitle method; however, it would allow me to eliminate the getEventsByTitle function by incorporating the query as an api parameter in both the getEvents and getEventsFullListing functions.
* Both methods will maintain  using separate api calls due to the fulllisting api requiring significantly more resources to gather the necessary data allowing us to use the fulllisting api on an as needed basis.
* If it is required for a page to display multiple types of events, instead of passing in a pipe delimited list to the filterTitle json key, we would need to add multiple modular events components passing in the filterTitle key with one string for each component.
* The filterTitle JSON key will still be utilized and  passed as an optional parameter in the getEvents and getEventsFullListing functions.
* Note: When the  filterTitle key is utilized in the json array, the query may return a larger selection of events due to the query searching multiple keys instead of only the 'title' key it will now return selections based on seraching 'title', 'title_external', and 'location' keys.

@nickdenardis , @breakdancingcat

breakdancingcat:
@AmbyrElan and @nickdenardis 

I'm on board with a couple of notes. 

1. I recommend modifying the proposed 'getEventsByTitle' to become instead 'filterEventsByTitle' and be used inside of either 'getEvents' or 'getEventsFullListing'. 

2. Passing in the raw component JSON as such:
public->function->getEvents(`$site_id`, `$limit = 4`, `$filter`)
public->function->getEventsFullListing(`$site_id`, `$limit = 4`, `$filter`)

3. Review the raw return from both methods to check if we can filter them the same way or need a conditional. 

4. Perform any additional modification of whatever is passed from the component to `$filter` in the EventsRepository, not in the ModularRepository (like determining if we're looking at title, title external, location should be done in the EventsRepository). 

**Note 1:** If `$filter` is not empty, `getEvents` limit can be 50 to run the query and then reduced by ->take($limit) after it's grouped by date. `getEventsFullListing` already does this so no changes there. 

**Note 2:** Ignore or remove `$end_date` in getEventsFullListing params for now, we have not built that out yet.

### To sum it all up:
'filterEventsByTitle' would accept a raw return of events, perform its filtering, and return the filtered events, to then be further modified in either 'getEvents' and 'getEventsFullListing' functions (to assign images or group by date). 

### Next steps
I would like to see a working example of these changes with just title passed in (not the other external_title or location at this time) before moving forward.

app/Repositories/EventRepository.php
```diff
@@ -87,6 +88,7 @@ public function getEventsFullListing($site_id, $limit = 4)
                         return $event;
                     })
                     ->take($limit)
+                    ->groupBy('date')
```
breakdancingcat, line 91: We do not want ot group the full listing by date, please remove

resources/views/components/events-featured-row.blade.php
breakdancingcat, line 1: This file will need to be reverted since we are not grouping fulllisting events by date

resources/views/components/events-featured-column.blade.php
breakdancingcat, line 1: This file will need to be reverted since we are not grouping fulllisting events by date

app/Repositories/ModularPageRepository.php
breakdancingcat, line 1: This should resolve itself when you merge develop back into your branch or rebase

app/Repositories/EventRepository.php
AmbyrElan: > We do not want to group the full listing by date, please remove

I originally took the approach of adding to each component without adding the grouped by date collection filter in an earlier commit ([feat: filtered events for all events components](https://github.com/waynestate/base-site/pull/767/commits/b9791188790e2d626219d3918b5c768b1e32b32c)) and found that grouping by date allowed me to significantly reduce the code within the individual components since I no longer required a condition to check for filtered events in this later commit ([refactor: events returns to group by date eliminating the need for a filter key check helping to minimize the code for each component](https://github.com/waynestate/base-site/pull/767/commits/f4c617302da55dceb5376a51a3ed6ef2aa3460de)).    

 I was curious what if it would throw anything else off by ordering the fulleventslisting by date?

Fw7424:
I agree with the idea of adding the function filterEvents into the existing Event functions. I'm not clear on the reason for added the group by date on the filter by title, does the make sure that only the first time that event shows up is the only that appears?

app/Repositories/EventRepository.php
AmbyrElan: I'd like to demo the current changes in our dev meeting as well to go over all the feedback.  The change I made allowed me to reduce looping on the front end code and have a consistent output based on checking how the data is sent from the API.  My goal was to reduce the code and I was conscientious in trying to make the output consistent and reliable.

AmbyrElan:
> I agree with the idea of adding the function filterEvents into the existing Event functions. I'm not clear on the reason for added the group by date on the filter by title, does the make sure that only the first time that event shows up is the only that appears?

I added the group by date based on how the output was keyed from the featured events. Grouping by date created consistent output to work with so that I was able to reduce the component code to work consistently whether the events were filtered or not.  I plan to demo in our dev meeting while we go over the comments on this PR.

PR #773, opened 2025-01-14
feat: use subsite-folder to determine how to build profile link.

Fw7424:
creating this as an alternative option to simply removing the first slash. https://github.com/waynestate/base-site/pull/771

Realized that after looking into this longer and at a few other situations that checking for subsite-folder, is a better way of handling there profile links on subsite and nested urls, without the need to create addtional pages. 

During an upgrade from Base upgrade from 4 to 5, when the [profile link creation was moved to the repository](https://github.com/waynestate/base-site/commit/b01c60c16c138df80efda7008c87b17a336921c5#diff-ab5a25447b98b9683b19e8a6e86a3c06eb033d2277d198cdc0b11ca425a32fe7R55), The a slash was added to the front of profile/.

This corrects this issue.

### DIFF
--- a/app/Http/Controllers/DirectoryController.php
+++ b/app/Http/Controllers/DirectoryController.php
@@ -34,9 +34,9 @@ class DirectoryController extends Controller
         $site_id = $this->profile->getSiteID($request->data['base']);
 
         if (!empty(config('profile.group_id'))) {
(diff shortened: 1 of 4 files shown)


chrispelzer (approved):
Yea this is the way I prefer it, keeps the subsite structure intack and doesn't rely on the relativeness of the url

nickdenardis (approved):
This approach is more consistent with other areas of Base. 👍

PR #774, opened 2025-01-16
Hotfix 8.13.4 - remove environment check and allow trustProxies for all

chrispelzer:
@nickdenardis suggested removing the `env()` check due to the increased resources. Since we were running the TrustedProxies middleware previously without an environment check, we can remove it.
https://github.com/waynestate/base-site/blob/8.10.6/app/Http/Middleware/TrustProxies.php
https://github.com/waynestate/base-site/blob/8.10.6/app/Http/Kernel.php#L21

### DIFF
--- a/app/Http/Controllers/ArticleController.php
+++ b/app/Http/Controllers/ArticleController.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
 * Status: Private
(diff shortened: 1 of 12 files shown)


nickdenardis (commented):
👍

PR #775, opened 2025-01-29
Fix: Accordion cutting off content

breakdancingcat:
Because of how browsers handle margins (margins of adjacent elements overlap) when the accordion.js is calcuating the inner height, it is not accounting for the top margin of the first heading/element.

This change removes the top margin of any first element within the "fold". 

| Before      | After      |
| ------------- | ------------- |
|image|image|


nickdenardis (commented):
👍

PR #776, opened 2025-02-04
refactor: convert `//` incompatible comments to be `/* */` proper css comments

chrispelzer:
Comments in CSS can only be `/* */` syntax
https://developer.mozilla.org/en-US/docs/Web/CSS/Comments

This cleans up any `//` comments and allows `make stylelintfix` run without any odd results


breakdancingcat (approved):

nickdenardis (commented):
👍

PR #780, opened 2025-03-03
refactor: upgrade nodejs version to be v22

chrispelzer:
Node v20 LTS - currently in maintenance till April 2026
Node v22 LTS - currently active and will go to maintenance October 2025
https://nodejs.org/en/about/previous-releases#release-schedule

No real other necessary changes for compatibility


nickdenardis (commented):
👍

PR #784, opened 2025-04-03
fix: remove CACHE_DRIVER as it isn't used

chrispelzer:
CACHE_DRIVER is not used anywhere in the codebase, only CACHE_STORE is


nickdenardis (commented):
👍

PR #791, opened 2025-06-02
FEATURE | Modular columns | Expanded classes from page fields, added new template guide page

breakdancingcat:
# Refining and documenting modular columns
- **New Template Guide:** added a new styleguide page to explain how to set up columns and use gutters.
- **Backgrounds:** Added a class to the main layout allowing the layout type (menu or no menu) to dictate if a component background will be edge-to-edge (full width) or contained (columns). 
- **Classes from page fields:** Modified the modular repository to group classes from page fields by which container they need to be applied to. This allows us to control how the aforementioned component backgrounds will behave. 
    - **"containerClass"** is applied to the `section`; modifying the padding to this container is prohibited, it would disrupt the max width of the container and disrupt column sizing. Top and bottom margins can be applied to modify spacing between other components when needed. (This is handled automatically in the modular repository.) 
    - **"backgroundClass"** is applied to a specific div just for the background. It gets positioned absolutely for the fullwidth template to create the edge-to-edge look, otherwise it sits inside the component's content boundary. 
    - **"componentClass"** are the rest of the classes that the user added that weren't filtered out for the container or background needs.
- **New gutter classes:** Added `gutter` classes to simplify adding padding/margin to components. 
    - Tailwind's spacing increments are too minimal for our general needs, so rather than picking from fractions of an em (`p-4` = 1rem, `p-8` = 2rem etc), there is `px-gutter-[sm-xl]` (or `mx-gutter-[sm-xl]`). Each jump from `lg` -> `xl` is 1rem and is easy to keep track of. Safelisting a bunch of padding/margin classes would bloat the css more than a few gutter classes.

## How classes get transformed from page field to component
<table>
<tr>
<th>Component config</th>
<th>HTML result</th>
</tr>
<tr>
<td>
Contained element

```
{
"id":0000,
"columnSpan":10,
"classes":"end bg-gold-50 px-gutter pb-gutter"
}

```

</td>
<td>

```
<section id="promo-row" class="mt:colspan-10 end"> 
<div class="component__container px-gutter pb-gutter bg-gold-50">

```

</td>
</tr>
<tr>
<td>
Full width element

```
{
"id":0000,
"backgroundImageURL":"/image.jpg",
"classes":"mb-gutter"
}

```

</td>
<td>

```

<section id="button-row" class="mb-gutter">
<div class="component__background bg-cover bg-top" 
     style="background-image:url('/image.jpg');"></div>
<div class="component__container">

```

</td>
</tr>
</table>

## Preview
image

### Accordion expanded:
image

## Notes
- I did not spend a lot of time writing words in the template guide to explain things, but I wanted a bare minimum of information at least devs can use until better instructions are written. (Volunteers welcome!) 
- I suspect there are areas I have done something weird and I am looking forward to reading creative feedback, other ideas, and learning any best practices I may have missed.  
- The template guide example columns are very tall, they will not stay that tall but y'know, baby steps. Something something heading margins. 
- I don't know where to put information on gutters. It is most relevant on the template guide but unsure if it deserves it's own page. 
- The location of the template guide in the menu is not my favorite. I am thinking templates could come before components in the menu, so you understand what components are doing before you go picking them. But since the component guide and the template guide are closely related, it feels awkward they have different parent menus. 
- The (pale) gold background's purpose is to show the column boundary, please do not let people just start adding background colors all over the place for fun. 

## Upcoming
Most of this list is already written either in other branches or on another project and is on it's way. 
- **Backwards compatibility:** Assign a default columnSpan:6 to any component with "column" in it's name. 
- **Pick a standard:** Add "classes" and/or "sectionClass" to the component guide. 
- **More full-width examples:** Modify the full width styleguide page with a few more typical examples (3 column and 4 column situations).
- **Easier styleguide components:** Bypass the need to make components in the controller or repository, components can be added to the styleguide page just like they are in the CMS, they run thru the real modular repository, and factory data is automatically generated. 
     - Rename factories to align with the component filename, this enables us to only need to set up the component in the page.
- **Modular layout config:** Add a new component called "modular-layout-config" that will allow you to 
    - Show/hide the page menu
    - Show/hide the page title
    - Show/hide page content
    - Set a class for the whole page 
- **Page content and title as a component:** Moves cms page content/page title into a component so we have less places with conditionals for the page title modifications.  

_The universe must have added --verbose when I was created._

### DIFF
--- a/app/Repositories/ModularPageRepository.php
+++ b/app/Repositories/ModularPageRepository.php
@@ -338,13 +338,13 @@ class ModularPageRepository implements ModularPageRepositoryContract
                 && !Str::contains($componentName, 'heading')
                 && $componentName != array_key_last($components)
             ) {
-                $components[$componentName]['component']['containerClass'] []= 'mb-gutter';
+                $components[$componentName]['component']['containerClass'] [] = 'mb-gutter';
             }
 
             // Implode party, assign all classes to their respective container
-            $components[$componentName]['component']['containerClass'] = implode(' ',$components[$componentName]['component']['containerClass']);
-            $components[$componentName]['component']['backgroundClass'] = implode(' ',$components[$componentName]['component']['backgroundClass']);
-            $components[$componentName]['component']['componentClass'] = implode(' ',$components[$componentName]['component']['componentClass']);
+            $components[$componentName]['component']['containerClass'] = implode(' ', $components[$componentName]['component']['containerClass']);
+            $components[$componentName]['component']['backgroundClass'] = implode(' ', $components[$componentName]['component']['backgroundClass']);
+            $components[$componentName]['component']['componentClass'] = implode(' ', $components[$componentName]['component']['componentClass']);
         }
 
         return $components;
--- a/styleguide/Http/Controllers/TemplateGuideController.php
+++ b/styleguide/Http/Controllers/TemplateGuideController.php
@@ -49,7 +49,7 @@ class TemplateGuideController extends Controller
 
 
             // ------------------------------------
-            // Column span 
+            // Column span
             // ------------------------------------
             'promo-row-1000' => [
                 'data' => app(EmptyPromo::class)->create(1, false, [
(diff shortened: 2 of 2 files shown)


nickdenardis:
@breakdancingcat 
This looks good to me! 👍 Thanks so much for your work on this 😊

KyleApsey (approved):
These updates look good to me!

PR #792, opened 2025-06-06
FEATURE | Column backward-compatibility | Support existing sites that use component columns, introduce column offset

breakdancingcat:
## Goal
Maintain backward-compatibility for existing sites while allowing full-width components. 

## Context
When converting the component loop to display full-width components, grid was removed from the component loop container in favor of flex. We still need the ability to display multiple columns in one row, but without a row container keeping the components aligned, a new way to control column widths and offsets was invented. In order to maintain backward-compatibility with existing sites, a default column span of 6 is applied to any component with 'column' in it's name, otherwise a columnSpan can be set by the user.

Going forward, components no longer need to be named with 'row' or 'column'.

## Changes
- Updated the min-width for columns from 8.25% to 8.3335% after encountering a 6px column discrepency.
- Added columnSpan: 6 to any component with column in it's filename. 
- Updated component-based styleguide pages to send the fake components through componentClasses() and componentStlyes() functions which applies all necessary classes components need to display in the loop.
- Created the ability to offset columns which will allow default centered columns less than 12 columns wide to shift left or right to align with other components.
- Added a section to the Template Guide page briefly explaining offsets.

### Example: Expanded Template Guide
(I'll find a more subtle background image, or swap the svg colors to gold/lighter gold.)
image

### Example: Styleguide page with columns
image

### Example: Styleguide page with offset single column
image


resources/scss/partials/_component-loop.scss
```diff
@@ -1,7 +1,7 @@
 @layer base {
     :root {
         --column-max: 100px;
-        --column-min: 8.25%;
+        --column-min: 8.3335%;
```
KyleApsey, line 4: Curious if `calc(100% / 12);` works here on our setup with the same end result?

KyleApsey (approved):
This looks great

_(I can't even begin to tell you how much I love that we have dedicated styleguides with faked content, instead of manually having to create them for all new projects with hardcoded test content.)_

resources/scss/partials/_component-loop.scss
breakdancingcat: It does! Thank you, I've made this change.

PR #793, opened 2025-06-06
Feature | .env | All site-specific ENV values now moved to the top of the file

nickdenardis:
## Reason for change

Because Laravel 11 requires more items to be defined directly in the `.env` file, the file has a mixture of default values for Laravel and unique values to each site.

## This change

Moves all the items specific to the site up top and separates the default values below.

This will allow for quick scanning and customization of a site without having to sift through every field in the file.

## Result

No visible change, not value changes, only re-organization.

## Schedule

This is scheduled for the 8.15 release. Hold merging until 8.14 is complete.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,11 +1,23 @@
+# Site specific configuration
 APP_NAME=base
 APP_ENV=local
 APP_KEY=
 APP_DEBUG=true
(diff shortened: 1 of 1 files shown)


breakdancingcat (approved):
Yay comments! Thanks!

PR #795, opened 2025-06-11
Feature | Config | Allow local config override through .env values

nickdenardis:
Bringing the ability to define site config values from a local file.

This allows the same code to be deployed to multiple sites where only the configuration values are different. Since the config file is a nested set of complex key => values, creating `.env` values for every one of them is not possible.

A new `CONFIG_OVERRIDE` field is defined in the `.env` file, which points to a file to replace any `config()` value. Because the config is cached after deploy, this update does not impact performance.

## Example override

```
# .env

APP_URL=https://somesite.wayne.localhost
CONFIG_OVERRIDE=/directory/to/override.config.php
APP_LOCALE=en
```

```
# override.config.php

<?php
return [
    "gtm_code" => "GTM-OVERRIDE",
    "layout" => "custom",
    "top_menu_enabled" => true,
    "new_config_key" => "test_value",
];
```

| Before                                                                                            | After                                                                                                                                    |
| ------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------- |
| ...<br>"gtm_code" => "GTM-NCBVKQ2",<br>"layout" => "main",<br>"top_menu_enabled" => false,<br>... | ...<br>"gtm_code" => "GTM-OVERRIDE",<br>"layout" => "custom",<br>"top_menu_enabled" => true,<br>"new_config_key" => "test_value",<br>... |

## Custom replacement function

A custom recursive replacement function was required to allow for non-associative arrays to be replaced wholesale. Alternatively, the native `array_replace_recursive()` function would only replace individual items from the original array with their indexed numbered counterparts from the new array.

## Cleanup

- This also cleans up when the `.env` file is copied from the `.env.example` to ensure it only happens on initial composer install
- The `APP_NAME` is used in the deployment to allow the deploy paths to be relative to the config values
- The hardcoded ID's are now removed from the promotion configuration

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,6 +6,8 @@ APP_TIMEZONE=America/Detroit
 APP_LOG_LEVEL=debug
 APP_URL=http://base.wayne.localhost
 
+CONFIG_OVERRIDE=
+
 APP_LOCALE=en
 APP_FALLBACK_LOCALE=en
 APP_FAKER_LOCALE=en_US
--- a/Envoy.blade.php
+++ b/Envoy.blade.php
@@ -3,9 +3,15 @@
 use Symfony\Component\Process\Process;
 
 /**
+ * Pull in environment
+ */
+$dotenv = Dotenv\Dotenv::createImmutable(__DIR__);
+$dotenv->load();
+
+/**
  * Application name
  */
-$appname = ''; // domain
+$appname = env('APP_NAME');
 
 /**
  * Remote server connection string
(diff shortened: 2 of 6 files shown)


nickdenardis:
This was merged into `develop` but done through merge conflict resolution commit:
https://github.com/waynestate/base-site/commit/7bb17f0062c6bfe35697141a33d7d2a75ec73f8f

It looks like that skipped detecting that this PR was open.

Closing.

PR #796, opened 2025-06-13
FEATURE | Layout config | Toggles to hide menu, page title

breakdancingcat:
## New component: layout-config
**Layout-config:** Use page field `modular-layout-config` with `showSiteMenu: false` and `showPageTitle: false` to hide these elements. Hiding the site menu will display the full-width template with full-width components.  

**Homepage controller:** Will display typical components when there aren't any components defined from the CMS.

**Page title:** Added display toggle inside partial. 

**Left menu:** Moved to a partial for a cleaner main layout file. 

**Code style:** Started reverting camel case php variables in favor of our typical snake case style.

-----
**Feedback request:** 
Since is the existing "layout" variable we use to set contained/main layout, I am looking for feedback if we should call this "template-config" instead.

**Future planning:**
I spent a lot of time thinking about where layout-config should go. It's currently nestled into 2 places, `MenuRepository` to be able to override `show_site_menu` and `PromoRepository` to take it out of the loop and attach it to the base object. It does not belong in either of these spots. 

Layout config needs to be set up before the menu repository is executed, and after discussing with Chris, we will explore adding to the data middleware to parse through cms page field data and format it to where it needs to go/what it will be used for. Please let me know if you have other things we could consider for phase 2 of this component.


nickdenardis (approved):
Looks good to me! Thanks for the video walkthrough 👍

resources/views/partials/page-title.blade.php
```diff
@@ -0,0 +1,8 @@
+{{--
+    // Replace title
+    'title' => $article['data']['title']
+
+    // Append string of class names
+    'class' => 'text-green divider-gold' 
+--}}
+<h1 @class([$class ?? '', 'visually-hidden' => (isset($base['page_config']['showPageTitle']) ? !$base['page_config']['showPageTitle'] : false)])>{{ $title ?? $base['page']['title'] }}</h1>
```
Fw7424: Wondering about why passing in $title and $base['page']['title']

Fw7424 (commented):
looks good to me.

breakdancingcat:
> Also though, Sounds like we're moving the "template-config" to be processed sooner than the rest of the modular components? If so I like that, and am wondering will we still need worry about how it gets handled later in modular other than maybe dropping.

@Fw7424 Hmm, it depends where all the configurations will need to happen. The modular repository is ensuring the json data gets assigned in the correct format, and the promo repository is removing it from the list of components to loop through. Ideally we would have the component data from page fields handled before the menu and promo repository so the logic isn't spread out through the app, but we can't/shouldn't do that until we rework the app and how the callbacks are handled.

resources/views/partials/page-title.blade.php
breakdancingcat: News articles and promo view need to change the page title. We are passing it in through the template for now.

PR #798, opened 2025-06-20
Fix | Code style | Replaced ternary operators in blade templates

nickdenardis:
## Reason for change

Consistency throughout the code.

```
<{{ !empty($item['link']) ? 'a href='.$item['link'] : 'div' }} class="{{ !empty($item['link']) ? 'group' : '' }}">
```
vs
```
@if (!empty($item['link']))  
  <a href="{{ $item['link'] }}" class="group">  
@else  
  <div>  
@endif
 ```

The second is preferred. It is easier to read and ensures HTML cannot be injected into the element. It is also consistent with the other conditionals thoughout the code.

## Changes 

- Replaced ternary operators in the middle of tags with expanded conditionals
- Updated the older array() style in blade templates

## Reminders

- Styleguide example in place

## Demo

_No visual changes_

### DIFF
--- a/resources/views/components/hero/half.blade.php
+++ b/resources/views/components/hero/half.blade.php
@@ -19,7 +19,7 @@
             <div class="hero__description text-lg 3xl:text-xl text-green-700 content">
                 @if(!empty($hero['description']))
                     @if(!empty($hero['link']))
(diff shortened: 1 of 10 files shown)


chrispelzer (approved):
This works out well, been trying to replace them where I've seen them at times.

PR #804, opened 2025-06-27
Fix | Styleguide accordions displaying configuration data again

breakdancingcat:
In an effort to keep our code style consistent, I updated the styleguide accordion variables to use snake case. This PR brings all of the styleguide controllers up to date so the configuration data displays once again. 

image


nickdenardis (commented):
👍

PR #805, opened 2025-06-27
Fix | Hero | Styleguide hero was overwritten by potential component hero

breakdancingcat:
Styleguide hero was overwritten by potential component hero. Assigned separate variables temporarily while I reevaluate how the promo repository and the styleguide promo repository should interact. 
image


KyleApsey (commented):
This looks good to me!

PR #809, opened 2025-07-07
Hotfix 8.14.3 | Fix yarn updating package.json on CI

chrispelzer:
Due to yarn on CI having the possibility to get updated, we could potentially have an error due to the package.json having a later yarn version than what was initially committed causing a git pull error. This will allow yarn to using the same version as in the package.json packageManager and allowing us to update as needed


nickdenardis (commented):
👍

PR #811, opened 2025-07-14
Feature | Hero | Contained available via option, updates to Carousel, Hero buttons

breakdancingcat:
## Reason for change
We were encountering issues with the `contained-hero` layout missing updates that were in the main layout file. Now there is only one layout file, the `contained-hero` layout includes main instead of repeating code. Because many of the shared CMS sites have set the layout to `contained-hero`, I have not removed the layout file completely until we can determine the best path forward without needing to update many config files.
 
Setting the layout to `contained-hero` will set the default hero to display contained, setting the layout to `main` will set the default hero to display banner small. Users can set any hero to contained or small using the dropdown options regardless of the layout setting. 

### Detailed changes
* New feature: Hero buttons available as a component.
* New feature: Hero contained option available on any layout.
* Updated Hero Carousel: Needs to be enabled with the `modular-hero` component to override the hero limit set in the base config.
* Moving in-line tailwind styles into hero.css for easier readability.
* Moved hero logic into it's own repository.
* Main and contained-hero layouts are exactly the same and will be phased out soon once I solidify how we will be swapping between contained hero and the other variations.
* Fixed the misaligned carousel drop shadow.
* Styleguide hero images load too slowly before flickity can determine the height, so it will resize again after the image is loaded.
* Hero components have been updated to use component data structure, allows us to now use the component config.
* Created test pages on base.wayne.edu to verify with live data.
    * [Hero default](https://base.wayne.edu/hero/default)
    * [Hero buttons](https://base.wayne.edu/hero/buttons)
    * [Contained](https://base.wayne.edu/hero/contained)
    * [Config override global hero](https://base.wayne.edu/hero/config-override)
    * [Config only](https://base.wayne.edu/hero/configonly)
* Created new styleguide page.
    * [Contained](https://base.wayne.edu/styleguide/component/hero/bannercontained)
* Updated carousel styleguide page with different instructions.
    * [Carousel](https://base.wayne.edu/styleguide/component/hero/carousel)

### Buttons option added into styleguide
image

### Contained option added into styleguide under components
image

### Carousel button drop shadow fixed, hover style added
half


styleguide/Repositories/PromoRepository.php
```diff
@@ -111,7 +109,21 @@ public function getRequestData(array $data)
...
         // Only pull hero promos if they match the page ids that are specificed
-        $styleguide_hero = !empty($hero_page_ids[$data['page']['id']]) ? $hero_page_ids[$data['page']['id']] : null;
+        $styleguide_hero = [
+            'data' => !empty($hero_page_ids[$data['page']['id']]) ? $hero_page_ids[$data['page']['id']] : null,
+            'component' => [
+                //'option' => '',
+            ]
+        ];
+
+        // TODO This is weak
```
nickdenardis, line 119: What does this mean weak? What edge cases should we also consider to make it stronger?

styleguide/Repositories/HeroRepository.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Styleguide\Repositories;
+
+use App\Repositories\HeroRepository as Repository;
+
+class HeroRepository extends Repository
+{
+    /**
```
nickdenardis, line 9: Is this a placeholder for the future?

nickdenardis (approved):
Looks good to me 😊

KyleApsey (approved):

PR #815, opened 2025-07-28
Hotfix 8.14.7 | update php to be v8.2.20 for local environment compatibility

chrispelzer:
Update PHP from v8.2.7 to v8.2.20 for local environment compatibility across sites.


nickdenardis:
👍

PR #821, opened 2025-08-22
Fix | Profiles | Legacy support for profiles_by_accessid custom page field restored

nickdenardis:
## Reason for change

With the update to use the `profile.` config for all profile listing configuration and `profile_config` JSON array to override in custom page fields, the previous `profiles_by_accessid` custom page field stopped working.

This change restores that ability for all sites currently using it until we can convert them to use the `profile_config` JSON format and ensures backwards compatibility. 

## Reminders

- Styleguide example in place


KyleApsey (commented):
Looks good to me

PR #822, opened 2025-08-22
Fix | Profiles | Legacy support for the base.profile_back_url restored

nickdenardis:
## Reason for change

With the update to use the `profile. config` for all profile listing configuration and `profile-config` JSON array to override in custom page fields, the previous `base.profile_back_url` custom page field stopped working.

Because this functionality exists on many sites, this change restores the ability to use `base.profile_back_url` without affecting forward progress on the configurations.

## Reminders

- Styleguide example in place

## Demo

| Base Config | Frontend |
|--------|------|
| Screenshot 2025-08-22 at 7 11 17 AM | Screenshot 2025-08-22 at 7 17 02 AM |


KyleApsey (commented):
Looks good to me

PR #824, opened 2025-08-25
FIX | Events column component | Rewrite conditional in modular repository

breakdancingcat:
### Reason for Change
The events column's data was directed through the fulllisting api call incorrectly, causing a 500 wherever that component was used.

I updated the problem conditional to ensure the data goes through the correct api call. (My component naming convention for events did me no favors.)

I also noticed that the "events row" was wrapping to the next line and I limited the default number of events to 3.

### Demo
image


KyleApsey (approved):
This looks good to me

PR #825, opened 2025-08-25
Fix | Config | Profiles config returned back to singular base.php configuration for singular file overrides

nickdenardis:
## Reason for change

As we work through ensuring backwards compatibility of existing sites and configurations, the separate `profiles.` configuration introduced multiple spots where similar values could be set.

This change brings all the profile configuration back into the same file to reduce complexity and allow for overrides in a single spot.

The same configurations are available through the `base.profiles.*` prefix.

## Reminders

- Styleguide example in place


KyleApsey:
This looks good to me

PR #827, opened 2025-09-15
FIX | Pre-commit | Ensure PHPStan only runs on base repository

KyleApsey:
## Reason for change
 
- The pre-commit hook was attaching itself to all repositories that use **base**, triggering a bunch of errors on individual sites.  This goes against the intention of adding PHPStan/Larastan where we really only want it to run on **base** and then we resolve all issues there before releases.
- This update adds a conditional check to confirm which should only allow sites with the repository name `base-site` to run `make phpstandry` on pre-commit.
 
## Relevant links
 
- _*This change was brought up in a conversation with @Fw7424 and there is no Basecamp or TP3 issue_
 
## Reminders
 
- Styleguide example in place

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -134,22 +134,32 @@ else
     printf "phpunit tests have passed.\n"
 fi
 
-# PHPStan/Larastan analysis
-if [ "${docker_test}" = "true" ]; then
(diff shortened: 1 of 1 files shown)


nickdenardis:
👍

PR #831, opened 2025-10-23
Hotfix 8.14.17 | Prevent https from being forced when on dev environment

chrispelzer:
This adds back behavior that was lost while upgrading Laravel which forced https on all URLs generated. Though we want to be able to have http when on www-dev URLs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "base",
   "private": true,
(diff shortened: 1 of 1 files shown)


nickdenardis (approved):
Nice! 👍

KyleApsey (approved):
Nice and simple.  Well done.

PR #835, opened 2025-10-30
Hotfix | Left menu displaying again when top menu is present

breakdancingcat:
### Restored left menu conditional

I moved the left menu into a partial to clean up the main layout file. I appear to have removed a very important conditional that made sure the menu displayed when the top menu was enabled. I believe at the time I thought it was duplicating an existing check. Clearly not! Since Base doesn't have top menu enabled by default, I did not catch this until it was deployed. 

I am looking for ideas on testing menu display changes to avoid this in the future. 

---

### Demo / Context

- [Basecamp](https://3.basecamp.com/5750155/buckets/37389969/todos/9231190810)

| Before | After |
|----|----|
|image|image|
|image|image|

---

### Final Checklist

- [ ] I have added or updated relevant documentation
- [ ] I have added tests (if applicable)
- [ ] I have communicated with the team about necessary post-deploy actions

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "base",
   "private": true,
-  "version": "8.15.0",
+  "version": "8.15.1",
   "description": "",
   "scripts": {
     "dev": "npm run development",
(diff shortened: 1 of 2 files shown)


nickdenardis (commented):
Looks good to me! 👍 Thanks 😊

KyleApsey (commented):
Nice!  Always something small like that.

Fw7424 (commented):
looks good

PR #844, opened 2025-12-10
Fix | Tests | Config override tests no longer rely on specific initial config values

nickdenardis:
## Correcting assumptions made in test values

Tests now use what ever values are set in the local repository to compare against instead of the original Base config values.

---

## Reason for Change

Hardcoding the config values in tests works until they are overwritten for a specific site. 

When overwritten for a site, tests start to fail.

This corrects that issue and reads the original config, for the specific site first before testing the overrides and compares against what ever values are in the local codebase.

---

## Demo / Context

| Before | After |
|--------|------|
| Screenshot 2025-12-10 at 6 46 40 AM | Screenshot 2025-12-10 at 7 03 15 AM |

---

### Final Checklist

---

## Testing Notes

Run tests normally with `make runtests`

### DIFF
--- a/tests/Unit/Support/BaseConfigTest.php
+++ b/tests/Unit/Support/BaseConfigTest.php
@@ -30,34 +30,49 @@ final class BaseConfigTest extends TestCase
     #[Test]
     public function config_returns_global_config_when_no_override_is_set(): void
     {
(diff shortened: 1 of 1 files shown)


breakdancingcat:
Thanks for getting to this, it's super helpful!

PR #867, opened 2026-02-13
FEATURE | Sub-site - Meta Image | Allow sub-sites to set sub-site specific default meta image

KyleApsey:
## Reason for Change

Some sub-sites have different meta images than their parent site, but right now the only way to change this is to set a meta image on every single page.  We needed a way to allow sub-sites to set a default fallback meta image (and corresponding alt text) different from their parent site but that itself could be overwritten by page.

---

### Links
- Basecamp: [https://3.basecamp.com/5750155/buckets/37389969/todos/9547303577](https://3.basecamp.com/5750155/buckets/37389969/todos/9547303577)

---

### Final Checklist

### DIFF
--- a/app/Http/Middleware/Data.php
+++ b/app/Http/Middleware/Data.php
@@ -43,8 +43,12 @@ class Data
         }
 
         // Overrides for meta information
-        $data['meta']['image'] = ! empty($page['data']['meta_image']) ? $page['data']['meta_image'] : '';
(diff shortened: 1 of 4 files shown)


nickdenardis (approved):
Looks good to me 👍

breakdancingcat (approved):
Great work, thanks!

PR #912, opened 2026-03-23
Feature | Accessibility | Anchor elements with multiple sub-elements now have accessible names

nickdenardis:
## Reason for change

Sites using anchor elements with multiple child elements (headings, images, descriptions) cause run-on sentences for screenreader users as it reads the entire contents of the anchor.

There are CSS techniques to work around this, but in the meantime, let's correct this for all sites using the standard components.

---

Each anchor wrapping card content (image + title + excerpt/description) now has aria-labelledby pointing to the id on the title element inside it, so screen readers announce the link using only the visible title text rather than the full card content.

### ID strategy

- `promo-{promo_item_id}` - spotlights, icons, single/catalog promos
- `news-{id}`- news row and featured news column
- `event-{event_id}` - featured event components
- `profile-{AccessID}` - profile grid

The heading partial accepts an optional `headingId` override so the ID can be passed through to title elements rendered via that partial.

---

## Catalog (current left vs updated right)

Screenshot 2026-03-25 at 6 12 17 AM

## Events

Screenshot 2026-03-25 at 6 12 06 AM

## Icons

Screenshot 2026-03-25 at 6 12 27 AM

## Featured news

Screenshot 2026-03-25 at 6 12 38 AM

## Spotlight

Screenshot 2026-03-25 at 6 12 47 AM

## Profiles

Screenshot 2026-03-25 at 6 12 56 AM

---

## Checklist

---

## Testing

- [x] Mac, Safari & VoiceOver
- [ ] Windows, Edge & NVDA

### DIFF
--- a/resources/views/components/events-featured-column.blade.php
+++ b/resources/views/components/events-featured-column.blade.php
@@ -2,11 +2,12 @@
     @if(!empty($data))
         @foreach($data as $item)
             <li class="mb-6">
-                <a href="{{$item['url']}}" class="flex w-full group">
+                @php $titleId = 'event-' . $item['event_id']; @endphp
+                <a href="{{$item['url']}}" class="flex w-full group" aria-labelledby="{{ $titleId }}">
                     <div class="w-1/3 shrink-0">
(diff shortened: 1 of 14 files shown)


KyleApsey (approved):
This looks good to me!

breakdancingcat (approved):
When the other solution is ready, do you want me to leave this change in as well? 

Thanks!

PR #924, opened 2026-03-30
UPDATE | config/base.php | Hero contained defaults and tests

KyleApsey:
## Reason for Change

I made an assumption (#919) based on incoming support requests that `full-width` and `banner` hero is the expected default.  After further discussion with @nickdenardis I believe that this was premature.

### "Banner" type
The "Slim" type hero is a new feature, so it was correct to set the "banner" type hero; we'll be keeping that as the default

### "Contained" placement
The "Contained" placement is our default, and we'll update individual sites that expect a full-width hero as they are reported.  The problem with enforcing "full width" onto sites that currently assume "contained" is that their hero images may be sized incorrectly causing fuzzy/low resolution images to be present.

---

## Final Checklist

### DIFF
--- a/config/base.php
+++ b/config/base.php
@@ -55,7 +55,7 @@ $global_config = [
     | ];
     |
     */
-    'hero_placement' => 'full-width',
+    'hero_placement' => 'contained',
 
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


breakdancingcat (approved):
Thank you!

nickdenardis (approved):
Makes sense to me 👍

Fw7424:
Thanks for this

PR #925, opened 2026-03-30
Fix | Profiles | Profile group dropdown is now always visible if a group is selected

nickdenardis:
## Reason for Change

The previous code didn't take into account the selected state for a group of profiles.

The updated code now only hides the dropdown in these two cases:
- All profiles on a page are in a single group
- There is a `forced_profile_group_id` from the site or page config

## Final Checklist

---

## Demo

| Before | After |
|--------|------|
| before | after |


KyleApsey (approved):
This looks good to me, thank you!

breakdancingcat (approved):
Looks good, thanks!

Fw7424:
Thanks for this update.

PR #936, opened 2026-04-10
Feature | Remove no console from eslint to allow ease of development use

chrispelzer:
### Remove no console from eslint to allow ease of development use

Removed the no-console configuration from eslint, which was causing eslint errors being thrown when using locally while developing. We can remove these manually but eventually the strategy with using vitejs will allow it to be done on build.

---

### Reason for Change

Issues while developing locally and compiling the js will throw eslint errors even when we want to use the console within the code locally.

---

### Final Checklist

---


eslint.config.js
```diff
@@ -27,7 +27,6 @@ module.exports = [
         },
         rules: {
             ...js.configs.recommended.rules,
-            'no-console': 'error',
```
nickdenardis, line 30: Looks good to me 👍

PR #938, opened 2026-04-10
Fix | Add check to 404 on the ArticleController when we are on a page greater than 0 and there are no articles coming back from the News API. 

chrispelzer:
### Add check to 404 on the ArticleController when we are on a page greater than 0 and there are no articles coming back from the News API. 

This only happens if you specifically update the ?page= in the URL, the prev/next don't show if it's the last page

---

### Reason for Change

Bots will crawl by updating the ?page= which would show a blank listing, but still register as a valid page even though the ?page= is greater than it should be.

---

### Demo / Context

Before (?page=5 with only 2 pages) :
SCR-20260410-ojpf

After (?page=5 with only 2 pages):
SCR-20260410-ojmi

---

### Final Checklist

---

### DIFF
--- a/app/Http/Controllers/ArticleController.php
+++ b/app/Http/Controllers/ArticleController.php
@@ -52,6 +52,14 @@ class ArticleController extends Controller
 
         $articles = $this->article->listing($request->data['base']['site']['news']['application_id'], 25, $request->query('page'), ! empty($selected_topic['topic_id']) ? $selected_topic['topic_id'] : null);
 
+        // 404 if no articles are found and we are trying to page which doesn't exist
+        if (
+            $request->query('page') > 0 &&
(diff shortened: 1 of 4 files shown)


chrispelzer:
Still need to add a test coverage for this

chrispelzer:
Tests are added as well as spliting out the tests for no page to handle both cases where no page with more than the total number of articles or less than the total number of articles to show on the initial listing page

KyleApsey (approved):

nickdenardis (approved):
Looks good to me! 👍

chrispelzer:
One more test added `news_listing_with_page_and_no_articles_should_404` for the ArticleController 404 error coverage

breakdancingcat:
Thank you!

PR #939, opened 2026-04-10
Fix | Correct the canonical url when on the PromoController::show to include the slug and promo id to match the current URL structure of the page

chrispelzer:
### Correct the canonical url when on the PromoController::show to include the slug and promo id to match the current URL structure of the page

Correct the canonical url when on the PromoController::show to include the slug and promo id to match the current URL structure of the page.

---

### Reason for Change

It was just showing the page without the slug-id which is what the page actually was loading as

---

### Demo / Context

Before:
SCR-20260410-ovwf

After:
SCR-20260410-ovrg

---

### Final Checklist

---


breakdancingcat (approved):
Thanks!

PR #992, opened 2026-06-22
Fix | Artisan | base:module generation now creates a usable accordion to display config

nickdenardis:
When generating a new modular component, the styleguide config and promotion group accordions do not expand by default.

The stub file was using a different array key than expected and the content was not visible.

This change updates the `promo_item_id` from dashes to underscores to match the expectations of the template.

---

## Demo / Context

| Before | After |
|--------|------|
| accordion-broken | accordion-fixed |

---

## Final Checklist


breakdancingcat (approved):
Thank you

PR #1010, opened 2026-07-10
Feature | Components | Added 'visibility' option to show or hide modular components regardless if data is available

nickdenardis:
By default, custom page fields only show when they have data (from promotions, news, or events). This is a great standard practice, but there are times when we may want to disable a component for a short time or display a component that isn't connected to a datasource at all.

This change allows us to determine the visibility of a component based on the "visibility" config option. It is a new optional option available to all existing and future components.

It allows us to have a formal way of hiding a component instead of changing the custom page field or putting a "/" or somehow making the JSON invalid to force skipping over the component.

It allows us to introduce components that do not rely on promotion groups (JavaScript only, third-party sources, static HTML, etc).

## Default behavior

`visibility: data-only` - The default visibility is to only show a component if there is data. No need to list this, but added for completeness/documentation.

## Temporarily hide

`visibility: hide` - Allows the content owner to "hide" a component while keeping everything about its name, position and config intact.

## Show without data

`visibility: always` - Right now, the only way to show a component that isn't connected to a promotion is to add it to the code as an exception. Being able to show something all the time, reguarless if there is data, decouples the component from a collection.

## Demo / Context

| Config | Result |
|--------|------|
| Screenshot 2026-07-10 at 1 24 45 PM | Screenshot 2026-07-10 at 1 25 02 PM |

## Final Checklist

## Testing Notes

Tests added to confirm each visibility option:

Screenshot 2026-07-10 at 1 26 12 PM

### DIFF
--- a/tests/Unit/Repositories/ModularPageRepositoryTest.php
+++ b/tests/Unit/Repositories/ModularPageRepositoryTest.php
@@ -6,6 +6,8 @@ use Factories\Article;
 use Illuminate\Support\Str;
 use PHPUnit\Framework\Attributes\Test;
 use App\Repositories\ModularPageRepository;
+use Contracts\Repositories\ArticleRepositoryContract;
+use Contracts\Repositories\EventRepositoryContract;
 use Factories\Page;
 use Factories\GenericPromo;
 use Factories\PromoWithOptions;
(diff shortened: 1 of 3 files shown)


KyleApsey (approved):
Looks good to me

KyleApsey (approved):
This looks good to me

tests/Unit/Repositories/ModularPageRepositoryTest.php
```diff
@@ -730,4 +732,106 @@ public function replace_modular_page_relative_url_with_filename_if_on_base(): vo
...
+        $promo_group_id_accordion_always = 5432;
+        $promo_group_id_accordion_hide = 1111;
+        $promo_group_id_accordion_default = 2222;
+
+        $data = app(Page::class)->create(1, true, [
+            'page' => [
+                'controller' => 'ChildpageController',
+            ],
+            'data' => [
+                // 1. Custom component, no ID, visibility = always -> should display
```
breakdancingcat, line 749: @nickdenardis I appreciate these specific test parameter comments!

breakdancingcat (approved):
This looks good, thank you

PR #1011, opened 2026-07-13
Refactor | Modular component Tabs | Accessibility - Add ARIA roles and arrow key navigation

tomkrupka:
Accesibility Review goals

Add in ARIA for tabs: currently there is no designated role for tabs (role="tablist", role="tab", role="tabpanel"). Without this, each tab just reads out like a link and there is not context given to screen reader users for the changing content.

Each tab section should be navigable using the horizontal arrow keys: right now you have to use the tab key to get through the full list of tab options.

### Relevant links
- https://dequeuniversity.com/library/aria/tabpanel
- https://www.w3.org/WAI/ARIA/apg/patterns/tabs/#keyboardinteraction

### Final Checklist
- [ ] I have added or updated relevant documentation
- [ ] I have added tests (if applicable)
- [ ] I have communicated with the team about necessary post-deploy actions

### Screen Recording of use of arrow key navigation
https://www.loom.com/share/5f4a174405514db39ae435d7e661f22f

### DIFF
--- a/resources/views/components/tabs.blade.php
+++ b/resources/views/components/tabs.blade.php
@@ -5,14 +5,20 @@
     @include('components.accordion', ['data' => $data])
 
     <div class="tabs__area js-tabArea">
-        <ul class="tabs__nav js-tabNav">
+        <ul class="tabs__nav js-tabNav" role="tablist" aria-label="Promo Tabs">
             @foreach($data as $item)
                 <li @class([
                         'tabs__nav-item',
                         'lg:w-1/' . $loop->count,
                         'active' => $loop->first
-                    ])>
+                    ]) role="presentation">
+
                     <a class="tabs__nav-link"
+                       id="tab-link-{{ $item['promo_item_id'] }}"
+                       role="tab"
+                       aria-selected="{{ $loop->first ? 'true' : 'false' }}"
+                       aria-controls="tab-{{ $item['promo_item_id'] }}"
+                       tabindex="{{ $loop->first ? '0' : '-1' }}"
                        href="#definition-{{ $item['promo_item_id'] }}"
                        data-target="#tab-{{ $item['promo_item_id'] }}">
                         {{ $item['title'] }}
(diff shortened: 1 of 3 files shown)


resources/views/components/tabs.blade.php
```diff
@@ -5,14 +5,20 @@
     @include('components.accordion', ['data' => $data])
 
     <div class="tabs__area js-tabArea">
-        <ul class="tabs__nav js-tabNav">
+        <ul class="tabs__nav js-tabNav" role="tablist" aria-label="Promo Tabs">
```
nickdenardis: Thinking this hard-coded "Promo Tabs" would need to be replaced with something dynamic or at least remove the "Promo" word, which won't mean anything to the end-user. 

If it has to be something static, let's make it "Tab selection".

nickdenardis (approved):
Looks good to me. 👍

resources/views/components/tabs.blade.php
breakdancingcat: @nickdenardis The dynamic option could be the component heading if it's available. Alternatively, the promo group title is available but often has other identifiers in it like "Tabs - Student success" etc. Let me know your thoughts, it can happen in a different PR.

nickdenardis: I yield to Skyler based on how/when it is announced in a screen reader to ensure it has enough context. It may not need to be super specific, but it may need to be.

breakdancingcat:
@tomkrupka I see coveralls is having some kind of issue. This is good to merge in for Skyler's review.

breakdancingcat (approved):

resources/views/components/tabs.blade.php
tomkrupka: Thank you @nickdenardis on the catch!
Should I pull in the promo group title like this? (To keep it unique if more than one on the page)
`<ul class="tabs__nav js-tabNav" role="tablist" aria-label="{{ current($data)['group']['title'] }}">`

Let me know if there is a better way or I could use "Tab selection"

tomkrupka: Removed aria-label via Skyler's preference

KyleApsey:
Looks good to me!

PR #1013, opened 2026-07-13
FEATURE | Caching | Fallback cache fallback for failing API response

KyleApsey:
## Reason for Change
A recent DB outage caused the WSU API to hang, and with no stale fallback, every page request blocked up to 20s before failing, taking down all front-end pages.  This applies graceful-degradation behavior (explicitly, with tests) and adds a follow-up requirement: stale data shouldn't resurrect content that has since expired on its own terms.

## Relevant Links
- https://3.basecamp.com/5750155/buckets/37389969/card_tables/cards/9976847319#__recording_10091181394

---
  
### Screenshots
index
event listing
stale cache
EventRepositoryTest
Cache tests

---

  ### Final Checklist
  
---
  
### Manual outage simulation (local dev): 
1. Update the .env to a `TTL=15` to set a, temporary, 15 second cache
2. Open a page on a base site, using this branch, to store page and promo data in stale cache fallback
3. `docker stop wsu-api` (and/or `wsu-mysql`) after warming the cache
4. Tail the logs (`tail -f storage/logs/laravel.log`) for the site you're testing with
5. Refresh the pages you previously warmed the cache for
6. Confirm the page loads the same as when `wsu-api` and/or `wsu-mysql` were running before
7. Confirm you see `StaleCache: falling back to stale data for cache key [...] after a null API response` warnings for each promo, event, page data that is cached in the logs
8. Recover `wsu-api` (and/or `wsu-mysql`) in Docker
9. Confirm the pages are still showing as expected

## Post deploy steps
1. Add the `CACHE_STALE_TTL=` property and set an override value in seconds to the _.env_ 
  - *Current default used in _.env.example_ and _config/cache.php_ tested on my local dev is `604800`, which is 7 days

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,6 +15,7 @@ NEWS_API_CACHE=../storage/app/newsapi/
 NEWS_API_ENDPOINT=https://news.wayne.edu/api/v1/
 
 TTL=0
+CACHE_STALE_TTL=604800
 
 CONFIG_OVERRIDE=
 
--- a/app/Repositories/ArticleRepository.php
+++ b/app/Repositories/ArticleRepository.php
@@ -5,9 +5,12 @@ namespace App\Repositories;
 use Contracts\Repositories\ArticleRepositoryContract;
 use Illuminate\Cache\Repository;
 use Waynestate\Api\News;
+use App\Traits\StaleCache;
 
 class ArticleRepository implements ArticleRepositoryContract
 {
+    use StaleCache;
+
     /** @var News */
     protected $newsApi;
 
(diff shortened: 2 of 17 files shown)


breakdancingcat (approved):
I appreciate the refactoring as well, thank you!

nickdenardis (commented):
@KyleApsey 
This looks really good! 

To confirm, going forward, `$this->rememberWithFallback` is a drop-in replacement for `$this->cache->remember` and should be used moving forward?

If so, we'll need to update it in the `base:feature` stub:
https://github.com/waynestate/base-site/blob/f928c758d3be2957f9ee89b41d210d123f7d6d67/stubs/repository.stub#L49

KyleApsey:
> @KyleApsey This looks really good!
> 
> To confirm, going forward, `$this->rememberWithFallback` is a drop-in replacement for `$this->cache->remember` and should be used moving forward?
> 
> If so, we'll need to update it in the `base:feature` stub:
> 
> https://github.com/waynestate/base-site/blob/f928c758d3be2957f9ee89b41d210d123f7d6d67/stubs/repository.stub#L49

Thank you for bringing this to my attention.  I've updated the stub file to use the new `rememberWithFallback()` method here in #1015

PR #1042, opened 2026-08-06
Fix | Promo view | Placeholder image no longer shown if promo item does not contain the optional field

nickdenardis:
The individual promotion item view has an optional image:
https://base.wayne.edu/styleguide/view/title-12345#definition-promo-setup

If there is no uploaded image, a placeholder image is put in its place.

Since this is an optional field, the display of the image should be optional. This change removes the placeholder image when no image is available.

It also cleans up the potential of an empty `<div>` if there is no back button to be shown.

## Checklist

## Demo

| Before | After |
|--------|------|
| Screenshot 2026-08-06 at 4 05 46 PM | Screenshot 2026-08-06 at 4 05 54 PM |

### DIFF
--- a/resources/views/promo-view.blade.php
+++ b/resources/views/promo-view.blade.php
@@ -5,13 +5,11 @@
         @include('components.page-content')
     @endif
 
-    <div class="w-full md:w-1/2 lg:w-1/3 mt-6 my-4 md:ml-4 float-right">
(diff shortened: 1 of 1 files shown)


KyleApsey (approved):
Looks good to me.  Thank you for updating this.

PR #1056, opened 2026-09-02
Feature | Fall back to batch profile image endpoint for missing pictures

chrispelzer:
Fall back to batch profile image endpoint for missing pictures. Adds an optional fallback that batch-fetches profile pictures when they are missing from the primary listing/view API response

---

### Reason for Change

Some profiles returned by the listing and view endpoints come back without a `Picture` value. This left gaps in the UI wherever a profile image was expected. The fix adds a fallback that calls `profile.users.batch` to fill in any missing pictures, so profiles reliably have an image available.

The fallback is opt-in, gated behind `base.profile.use_global_image` (configurable via the `profile-config` custom field), so sites only get this extra request when they explicitly enable it.

---

### Demo / Context

- New config option: `base.profile.use_global_image` (see `config/base.php`)
- `ProfileRepository::hasProfilesMissingPictures()` checks a batch of profiles for missing pictures
- `ProfileRepository::getProfileImages()` performs the batch fallback request (cached via `rememberWithFallback`) and merges results back into the profile data
- Applies to both the profile listing path and the single-profile (`site_id`) lookup path

---

### Final Checklist

---

### Testing Notes

- New coverage added in `tests/Unit/Repositories/ProfileRepositoryTest.php`

---

### DIFF
--- a/app/Repositories/ProfileRepository.php
+++ b/app/Repositories/ProfileRepository.php
@@ -71,6 +71,11 @@ class ProfileRepository implements ProfileRepositoryContract
         // Make sure the return is an array
         $profiles['profiles'] = empty($profile_listing['error']) ? $profile_listing : [];
 
+        // Fall back to the batch faculty image endpoint for any profile missing a picture
+        if (config('base.profile.use_global_image') && !empty($profiles['profiles']) && $this->hasProfilesMissingPictures($profiles['profiles'])) {
+            $profiles = $this->getProfileImages($profiles);
+        }
+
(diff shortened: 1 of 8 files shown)


nickdenardis (changes requested):
@chrispelzer 
This looks really good!

Before approving it though, can you add the option to the styleguide documentation?
https://base.wayne.edu/styleguide/profiles

breakdancingcat (approved):
Thank you!

PR #1061, opened 2026-09-11
Fix | prevent stale .envoydeploy source from being silently redeployed

chrispelzer:
- Type: Bug fix
- Affects: Envoy deployment pipeline (`Envoy.blade.php`)
- Changes: Reset/clean the local `.envoydeploy` source clone before checkout, and add `set -e` to the local-source tasks so failures abort the deploy instead of continuing silently.

---

### Reason for Change

`updaterepo_localsrc`'s `git checkout -B` can fail when the local `.envoydeploy/source` clone has uncommitted changes left over from a previous build (e.g. `yarn.lock` rewritten by `make yarn`). Because no task aborted on failure, the pipeline would carry on and package/deploy whatever was still sitting in the stale clone, tagging the tarball with the current (correct) commit hash. Every later redeploy of that same commit would then reuse the mislabeled cached tarball, masking the problem.

---

### Demo / Context

- `git reset --hard HEAD` and `git clean -fd` are run against the local source clone before `git checkout -B` to discard any leftover changes from a previous build.
- `set -e` added to `updaterepo_localsrc`, `depsinstall_localsrc`, and `packrelease_localsrc` so a failure in any of these tasks stops the deploy instead of silently continuing with stale content.

---

### Final Checklist

---

### Testing Notes

- Verify a deploy still succeeds normally when the local source clone is clean.
- Verify that if `.envoydeploy/source` has uncommitted/stray changes (e.g. from a prior `make yarn` run), the deploy now discards them and proceeds with a clean checkout rather than silently packaging stale content.
- Verify that a genuine failure in `updaterepo_localsrc`, `depsinstall_localsrc`, or `packrelease_localsrc` now aborts the deploy instead of continuing.


KyleApsey (approved):

nickdenardis (approved):
👍

PR #1062, opened 2026-09-11
UPDATE | Component loop | Improve Sass approach

KyleApsey:
## Reason for Change
After discussing when to use CSS variables and when to use Sass variables and how we can leverage some of Sass's advanced features, @breakdancingcat suggested that I take a look at the `_component-loop.scss` which turns out is the perfect example of all of these things.

### When to use Sass variables instead of CSS variables
In Dart Sass, Sass variables are file scoped, unless they are imported or explicitly declared globally.  This means you can have for example a `$background-color` variable in each of your files and they never overwrite each other.  This is because fundamentally Sass variables are converted into their value at compilation, so:
```Sass
@use "../../utilities/colors.scss" as colors;

$background-color: color.scale(colors.$green-400, $lightness: -8%);

.my-selector {
  background-color: $background-color;
}
```
Is compiled to, and seen by the browser as:
```Sass
.my-selector {
  background-color: #07302a;
}
```

We generally want to use Sass variables, because they are (1)tighter scoped, (2)are compiled into a fixed value, and (3)are, arguably, cleaner and easier to work with in the _.scss_ file.

### When to use CSS variables instead of Sass variables
You'll notice the `--column-min` is still defined as a CSS variable.  The key reason why you'd want to use a CSS variable is when the value can't be set at compilation, ie. the calculation to evaluate the rendered value is dependent on the the device that it's being viewed on.  In the case of `--column-min`, we want to base the minimum depending on a calculation of the available space; which is only accurate in the browser.

Another case would be if we need JavaScript/PHP/user generated content to drive a CSS value that we won't have at compilation.  For example, defining offset for content based on a responsive header/element which requires JavaScript recalculation at different screen sizes or based on different user generated content that's only ever on the page (like a responsive field or textarea).

### `@for` loop in Sass
I was able to reduce the amount of lines for defining the dynamic column widths from 35 to 4, which besides subjectively cleaner, this reduces the risk of typos and makes future updates much simpler.  For example, if we wanted to enable 16 columns instead of 12, there are no new lines added and nothing to copy/paste, just simply update the `$column-count` variable to `16` and the `.colspan-#{$i}` selector increases to 16 and the calculations remain accurate up to 16 columns.
_*Link to Sass `@for` loop [documentation](https://sass-lang.com/documentation/at-rules/control/for/)_

---

### Relevant links
- Basecamp: 

### Screenshots
after - component loop

---

## Final Checklist

---


breakdancingcat (approved):
This is exactly what we need and provides a solid example for how to choose the type of variable you need. This is the first time we've had a for loop in our CSS! Thank you!

nickdenardis (approved):
👍

PR #1065, opened 2026-09-14
Feature | Spotlights | Spotlights now support an optional image

nickdenardis:
Over time, as more people use the "Spotlights" area, pulling from promotions works well.

In some cases, there needs to be a spotlight that doesn't contain an image.

There was no conditional logic around the image in the modular component, which results in a broken image if someone uses the component without an image.

## Change

This change makes the image optional so it can be used intentionally without an image (or accidentally) to ensure this component is as flexible as possible.

## Final Checklist

## Demo

| With images | Without images |
|--------|------|
| Screenshot 2026-09-14 at 1 18 48 PM | Screenshot 2026-09-14 at 1 18 57 PM |


KyleApsey (approved):
Looks good to me.

PR #1066, opened 2026-09-14
UPDATE | Global buttons | Improve Sass approach, breakout into utilities

KyleApsey:
## Reason for Change
Upon request from @breakdancingcat I've refactored the areas where we were previously using the hungry `@extend` with the more targeted `@include`.  At the same time, I created a dedicated `utilities/buttons.scss` to house the button specific mixins/helpers and `utilities/colors.scss` to store the color specific Sass variables exported from the Tailwind config.

### Why `@include` instead of `@extend`
`@extend` merges the selector with all internal selectors and all combinations, which is particularly bloated and unpredictable when combined with the utility class names/selectors from Tailwind.  What you end up with is cascading rules that contain the individual and combination of every single selector between the extended selector and the ones being `@apply`-ed.

`@include` instead copies the mixin's declarations directly into each rule, right where it's called.  It acts more like you are copy/pasting the styles of the mixin in-place of the `@include` statement, and is compiled into pure CSS instead of compiled into layered CSS rules.

### What are these new files?
I took this chance to build towards what an idiomatic Dart Sass version of **Base** might look like.  I followed the pattern that I've been using for the [v3 CMS styles](https://bitbucket.org/waynestate/cms/src/master/resources/css/v3/utilities/).

We were `@extend`-ing the `.button` selector inside of __slate.scss_, __formy.scss_, and __flag.scss_, so creating a `utilities/buttons.scss` that we can `@include` just where we need in the above files and __global-buttons.scss_ without any repeated code.

I also created a _utilities/colors.scss_ that uses the colors defined in the Tailwind config or that we were previously accessing with `@apply {tailwind-utility-class}`.  Same as the `@include` logic for the buttons mixins, the colors variables can be imported into any file we want to use them with the `@use`, as seen in _utilities/buttons.scss_.

---

### Relevant links
- Basecamp: https://app.basecamp.com/5750155/buckets/37389969/card_tables/cards/10295612898

### Screenshots
_*No visual changes, these are structural and semantic changes only_
| Before | After |
|--------|--------|
| before - buttons (top) before - buttons (bottom) | after - buttons (top) after - buttons (bottom) |

## Final Checklist

---


nickdenardis (approved):
👍

breakdancingcat (approved):
Aw yess, thank you!

PR #1067, opened 2026-09-15
Feature | Profiles | Profile field configuration moved to config and centralized field output

nickdenardis:
Customizing the fields under the profile listings requires editing the views. This is needed more often now as some sites want titles or research interests on the listing views, requiring a custom codebase for a few lines of edits. 

Additionally, because the codebase displays profile fields on various pages, the logic on how each field should be displayed is throughout multiple views.

To provide flexibility on the profile listing pages and consistency between the different views this change:
- Centralizes the profile field configuration alongside the other profile configs
- Centralizes the profile field views so they can be used anywhere consistently
- Allows the `profile-config` overrides on each page for further customization

Finally, this change also addresses the accessibility of the listing view, where the link to the profile encapsulates the entire block of text but only labels it as their name. This hides the title or any additional fields from being readable outside of the links. With the customization of fields under the name, this change makes that content accessible and linkable again.

## Final Checklist

| Before | After |
|--------|------|
| Screenshot 2026-09-15 at 10 04 01 AM | Screenshot 2026-09-15 at 10 03 43 AM |


KyleApsey (approved):
Looks good to me

resources/views/components/profile-field.blade.php
```diff
@@ -0,0 +1,53 @@
...
+    } elseif (in_array($field, $fileFields) || (is_array($data) && isset($data['url']))) {
+        $items = isset($data['url']) ? [$data] : (is_array($data) ? $data : []);
+    } elseif (is_array($data)) {
+        $items = $data;
+    } else {
+        $items = [$data];
+    }
+@endphp
+@foreach ($items as $item)
+    @if (!empty($item))
```
chrispelzer: Is there a Blade function that combines the `@foreach` and `!empty($item)`? Though I think it might check the `$items` and not singular `$item`

chrispelzer (approved):

resources/views/components/profile-field.blade.php
chrispelzer: or filter the empty $items in the `@php` above, which will element this need

chrispelzer (approved):

resources/views/components/profile-field.blade.php
breakdancingcat: @chrispelzer are you thinking `@forelse`?

resources/views/components/profile-field.blade.php
breakdancingcat, line 1: @nickdenardis Should we make a new card to integrate this into the repository instead?

app/Repositories/ProfileRepository.php
```diff
@@ -327,27 +327,27 @@ public function getFields()
     {
         return [
             // Show under the profile image
-            'contact_fields' => [
+            'contact_fields' => config('base.profile.contact_fields', [
```
breakdancingcat, line 330: @nickdenardis Can you let me know the benefit of keeping the fields here and in the config?

breakdancingcat (approved):
A couple questions but overall super happy for this feature, thank you!

resources/views/components/profile-field.blade.php
KyleApsey: `@forelse` checks that `!empty($items)`.  I don't know off the top of my head and couldn't find any sort of `@foreachpresentitem`.  You could short circuit like:

```php
@foreach ($items as $item)
    @continue (empty($item))
        [...]
@endforeach
```

_But that is code style and while I personally like short circuiting over nested conditionals (can't remember the CS name for that pattern) it's not like this is a deep difficult to read conditional._

KyleApsey (approved):

app/Repositories/ProfileRepository.php
nickdenardis: Staged approach. Another feature is coming to pull more into the config for the contact tables. Then a final step to consolidate it all. But it will be based on a few decisions that can't be made until the contact tables PR.

resources/views/components/profile-field.blade.php
nickdenardis: Thanks for the feedback. I'll explore a few options, the performance implications and their general readability before providing an update.

nickdenardis: Cleaned this up and pre-filtered the null values, which removed the need to check the value in the display loop 😎

Change: https://github.com/waynestate/base-site/pull/1067/changes/08c9a011bc1bb4c09726ed6c6142f6536c88a914

PR #1068, opened 2026-09-16
Feature | Profiles | Contact tables now have customizable fields and Name link

nickdenardis:
With the customization of the profile contact fields on the directory type listing pages, the next step is to customize the contact table lists.

This change tackles a few things:
1. Moves the hardcoded table columns from the view to the central profile config
2. Allows the name to link to one of three places: "Email", "Profile", "None"
3. Allows overrides of these values on a per-page config.

This is the first of a few changes. After this is possible, the profile listings should become true modular components and be usable on any page alongside other modular components.

## Possible questions

1. The config field `table_name_link` was a trade-off to allow three options instead of using the "singlePromoView" type approach with a `singleProfileView` option that only allows email or profile link.
2. Because this table is on the page multiple times, without specific column widths on all but one column, the tables are very hard to scan and read as the columns don't line up down the page. This code chooses a static width for the name and the third-to-last column.

## Final Checklist

## Tests

Screenshot 2026-09-16 at 9 14 17 AM

## Demo

| Before | After |
|--------|------|
| Screenshot 2026-09-16 at 9 15 24 AM | Screenshot 2026-09-16 at 9 15 13 AM |


breakdancingcat (approved):
Looks good, thank you!

PR #1069, opened 2026-09-16
Feature | Component | Listserv subscription row component added

nickdenardis:
Initial proof of concept to bring the listserv subscription form into a site. As more sites use this component, we should make it open to everyone consistently.

This change brings the component into a modular row with the option of connecting it to a promotion group.

It offers two variations:
1. Connected to a promotion group to pull in the title, description and optional image
2. Used on its own with a heading provided by the "heading" in the custom page field

This is a row only at the moment to get an initial proof of concept up before working through a row version.

Requires:
- "list" => "your-list-slug", // Required. Formy subscriptions list slug.
- "form_id" => "4970", // Required. Formy form id, e.g. from "form-4970".
- "sitekey" => "your-recaptcha-sitekey", // Google reCAPTCHA site key for the form.

Styleguide uses the public Google reCAPTCHA public key for demo purposes only.

## Final Checklist

## Tests

Screenshot 2026-09-16 at 2 27 43 PM

## Demo

Screenshot 2026-09-16 at 2 39 54 PM


resources/views/components/listserv-subscribe-row.blade.php
```diff
@@ -0,0 +1,85 @@
...
+--}}
+@if(!empty($component['list']) && !empty($component['form_id']) && !empty($component['sitekey']))
+    @php
+        $item = current($data) ?: [];
+        $formId = 'form-'.$component['form_id'];
+        $headingLevel = !empty($component['heading']) ? 'h3' : 'h2';
+        $headingText = $item['title'] ?? $component['formHeading'] ?? 'Subscribe to our mailing list';
+        $image = $item['relative_url'] ?? $component['image'] ?? null;
+        $imageAlt = $item['filename_alt_text'] ?? 'decorative';
+        $actionUrl = 'https://forms.wayne.edu/subscriptions/?'.http_build_query(array_merge(['list' => $component['list']], $component['campaign'] ?? []));
```
KyleApsey, line 29: I know this is a prototype, but just wondering if this is eventually going into a repository or if it is something that belongs here in the template.

KyleApsey (approved):
Looks good.  Just one non-block question.

resources/views/components/listserv-subscribe-row.blade.php
nickdenardis: @KyleApsey 
Since components don't have repositories, the view is the only place to validate and set defaults. 

Let's explore what repositories for components may look like in a future release.

This php block method up top at least separates a good chunk of logic into a single location instead of being littered throughout the view output.

KyleApsey: Makes sense to me, thank you for the explanation.  No need to rethink how we build components right now, especially if it's a consistent pattern

breakdancingcat: @nickdenardis @KyleApsey My goal is for components to have repositories, or at least types of components, to categorize logic outside modular repository. I would have put this logic in the modular repository, but since it is already so stuffy I am fine with this too. Thank you for bringing it up!

breakdancingcat (approved):
Thank you!

PR #1070, opened 2026-09-17
Fix | Correct article hero images to override the CMS hero images when found

chrispelzer:
Type: Bug fix
Affects: Individual Articles with Hero images from the Article
Changes: Correct overriding rather than appending to the Hero array as well as making sure heroPlacement is set if not found

---

### Reason for Change

The article hero image was being pushed onto hero.data, which only rendered if the page already had a CMS hero component (and even then appended as an extra rotating slide instead of replacing it). Now the article's hero image replaces hero.data and ensures heroPlacement is set so it always renders.

### Final Checklist

---


nickdenardis (approved):
Thanks for getting to the bottom of this. 👍

breakdancingcat (approved):
Wonderful, thank you!

PR #1071, opened 2026-09-18
Feature | Deployment | Composer optimizations, exclude files and file attributes from final tar

nickdenardis:
Over the last few weeks, while deploying individual sites and the cms-base sites, we noticed:

- We should always be shipping the production composer files to dev/prod. The "-dev" dependencies are for local development only
- When deploying locally, there are extra dot files included in the tar file related to Mac OS attributes
- When looking at the final resulting files in each directory, there are exact development files which are not needed to run the site.

This change ensures we are only building the files needed for the site to load and keeping all development files in our local environment.

It reduces the number of files we ship to each site, which cuts down on the deployment time. (1 meg savings may not seem like a lot, but multiplying it by 200+ sites compounds the impact).

## Final Checklist

## Demo

| Before | After |
|--------|------|
| Screenshot 2026-09-18 at 12 07 36 PM | Screenshot 2026-09-18 at 4 09 26 PM |

## Tests

Screenshot 2026-09-18 at 4 26 48 PM


KyleApsey (commented):
Looks good to me

